PR #2, opened 2018-12-24
Add FR/EN reset password emails

manuquentin:
MA-33: Password reset

- [x] Add templates for EN and FR
- [ ] Use dedicated template depending on the user language
- [ ] Check if `firstName` and `lastName` are filled in the template
- [ ] Check why `host` is not `demo.wazo.community`


manuquentin:
Closed in favor of #3

PR #3, opened 2019-02-19
Update reset password email that redirects to songbird

manuquentin:


pc-m:
Hi @manuquentin 

sorry about the very long delay before an answer.

Do you still have https://songbird.wazo.io/reset-password somewhere?

manuquentin:
Hi @pc-m !

I've updated the link, songbird use hash now so it's https://songbird.wazo.io/#reset-password

pc-m:
merged into master thanks.

PR #11, opened 2019-07-08
WAZO-889-integrate-microsoft-plugin

pc-m:


wazo_auth/plugins/external_auth/microsoft/http.py
```diff
@@ -0,0 +1,119 @@
...
+
+from .helpers import get_timestamp_expiration
+from .schemas import MicrosoftSchema
+from .websocket_oauth2 import WebSocketOAuth2
+
+logger = logging.getLogger(__name__)
+
+# Allow token scope to not match requested scope. (Requests-OAuthlib raises exception on scope mismatch by default.)
+os.environ['OAUTHLIB_RELAX_TOKEN_SCOPE'] = '1'
+os.environ['OAUTHLIB_IGNORE_SCOPE_CHANGE'] = '1'
```
sileht: sigh... unfortunately oauthlib allow to set this only via env, such a shame for a lib to have to choice the behavior with environment only...

wazo_auth/plugins/external_auth/microsoft/http.py
```diff
@@ -0,0 +1,119 @@
...
+        self.scope = config[self.auth_type]['scope']
+        self.token_url = config[self.auth_type]['token_url']
+        self.user_service = user_service
+        self.websocket_host = config[self.auth_type]['websocket_host']
+
+    @http.required_acl('auth.users.{user_uuid}.external.microsoft.create')
+    def post(self, user_uuid):
+        client_id, client_secret = self._get_external_config()
+        self.user_service.get_user(user_uuid)
+        self.oauth2 = OAuth2Session(client_id, scope=self.scope, redirect_uri=self.redirect_uri)
```
sileht: We would validate the user input and creds first before requesting the external_config and create the oauth2 session.

integration_tests/suite/test_auth_microsoft.py
```diff
@@ -0,0 +1,188 @@
...
+        response = requests.post(url, json=body, headers=headers, verify=False)
+        response.raise_for_status()
+
+        cls.client = Client(
+            'localhost',
+            port=port,
+            verify_certificate=False,
+            username=cls.username,
+            password=cls.password
+            )
```
sileht: flake8 will not be happy :)

wazo_auth/plugins/external_auth/microsoft/http.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def _get_external_config(self):
+        tenant = Tenant.autodetect()
+        config = self.external_auth_service.get_config(self.auth_type, tenant.uuid)
+
+        return config.get('client_id'), config.get('client_secret')
+
+    @staticmethod
+    def _create_get_response(data):
+        return MicrosoftSchema().dump(data)
```
sileht: Just my 2cents: not sure this method worse it, it's the same number of chars and doesn't improve readability

wazo_auth/plugins/external_auth/microsoft/http.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if args.get('scope'):
+            self.oauth2.scope = args.get('scope')
+
+        logger.debug('User(%s) is creating an authorize url for Microsoft', str(user_uuid))
+
+        authorization_url, state = self.oauth2.authorization_url(self.authorization_base_url)
+        logger.debug('Authorization url : {}'.format(authorization_url))
+
+        self.websocket = WebSocketOAuth2(
```
sileht: I'm really not fan of having a http endpoint spawning a thread. If one day we run the http side of dird into an other wsgi app/lib, this is not going to play well.

sileht:
Just some remarks, nothing to fix here.

wazo_auth/plugins/external_auth/microsoft/http.py
pc-m: agreed

wazo_auth/plugins/external_auth/microsoft/http.py
pc-m: I agree with you, the google plugin uses the same mechanism. I'm looking forward to change this implementation and merge the code from both plugins.

wazo_auth/plugins/external_auth/microsoft/http.py
```diff
@@ -39,17 +39,17 @@ def __init__(self, external_auth_service, user_service, config):
...
-        client_id, client_secret = self._get_external_config()
-        self.user_service.get_user(user_uuid)
-        self.oauth2 = OAuth2Session(client_id, scope=self.scope, redirect_uri=self.redirect_uri)
         args, errors = MicrosoftSchema().load(request.get_json())
-
         if errors:
             raise UserParamException.from_errors(errors)
 
         if args.get('scope'):
             self.oauth2.scope = args.get('scope')
```
sileht: This need to move after the oauth2 creation or:
Oauth2Session should take `scope=args.get('scope', self.scope)`

PR #36, opened 2019-10-02
Pull request for WAZO-1152-refresh-token

sileht:
Really clean token every minutes
Send event when token expire soon

Depends-On: https://github.com/wazo-platform/xivo-bus/pull/9


sileht:
recheck

PR #43, opened 2019-10-23
Pull request for WAZO-1250-https-removal

sileht:
config: move rest_api/https to rest_api
Fix bootstrap when ssl is disabled


sileht:
recheck

PR #44, opened 2019-11-04
Pull request for WAZO-1297-bootstrap-improvement-v2

sileht:
Rework wazo-auth-bootstrap process

Why:
 Currently the master password is auto generated by this script, a two files must be created on the node and wazo-auth must be up. This makes
 very complicated the bootstrap on multiple nodes or with readonly containers setup.

Currently the process is:

* run wazo-bootstrap setup
  * Will create the /var/lib/wazo-auth/init key file
  * enable init plugin
* restart wazo-auth
* run wazo-bootstrap complete
  * it generates a password
  * send /var/lib/wazo-auth/init and generated credentials to endpoint /0.1/init that will:
     * assert /var/lib/wazo-auth/init content is the key provided via API
     * Create the initial user and add the default policy
     * delete /var/lib/wazo-auth/init
  * create /root/.config/wazo-auth-cli/050-credentials.yml
  * delete /var/lib/wazo-auth/init
* restart wazo-auth

New process do:

* just run wazo-bootstrap complete
  (We don't care wazo-auth is running or not, but database must have been initiated)
  * if /root/.config/wazo-auth-cli/050-credentials.yml does not exists
    * it creates it with a new auto-generated password
    * otherwise it reuses the existing credentials
  * it will creates intial user and add the default policy directly to
    the existing database

OR:

* just run wazo-bootstrap initial-user  --password <initial_user_password>
  (Same as complete but without the /root/.config/wazo-auth-cli/050-credentials.yml creation/usage)

The wazo-bootstrap initial-user also have --database-uri, --username,
--policy-name to configure the initial user.

All options can be passed with environment (eg:
WAZO_AUTH_BOOTSTRAP_USERNAME, ...)

If you rerun this command with the same options, they will be
idempotent.

Also `wazo-bootstrap setup` is now a noop and kept for backward
compatibility.


sileht:
recheck

sileht:
https://github.com/wazo-platform/wazo-auth-client/pull/11
https://github.com/wazo-platform/wazo-auth-cli/pull/9

sileht:
recheck

sileht:
\o/

fblackburn1:
Hey guys, you cannot change the bootstrap procedure without changing the installation scripts. Please update the ansible for wazo and all other projects that use wazo-auth

sileht:
> Hey guys, you cannot change the bootstrap procedure without changing the installation scripts. Please update the ansible for wazo and all other projects that use wazo-auth

I should not break anything, "wazo-auth-bootstrap setup" still exist and do nothing. And "wazo-auth-bootstrap complete" works as before.

fblackburn1:
> > Hey guys, you cannot change the bootstrap procedure without changing the installation scripts. Please update the ansible for wazo and all other projects that use wazo-auth
> 
> I should not break anything, "wazo-auth-bootstrap setup" still exist and do nothing. And "wazo-auth-bootstrap complete" works as before.

Yeah sorry you right, other failures misleads me

PR #47, opened 2019-11-07
allow user to manage his sessions

fblackburn1:
@pc-m I don't know what is the right format since we have a lot a different acl format
`auth.users.{{ user_uuid }}`
`auth.users.{{ uuid }}`
`auth.users.me`

### DIFF
--- a/alembic/versions/2e9c110c81aa_add_session_acl.py
+++ b/alembic/versions/2e9c110c81aa_add_session_acl.py
@@ -1,0 +1,117 @@
+"""add_session_acl
+
+Revision ID: 2e9c110c81aa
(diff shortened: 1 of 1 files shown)


pc-m:
> @pc-m I don't know what is the right format since we have a lot a different acl format
> `auth.users.{{ user_uuid }}`
> `auth.users.{{ uuid }}`
> `auth.users.me`

For the refresh token feature I used `{{ user_uuid }}` and `me`

fblackburn1:
> For the refresh token feature I used `{{ user_uuid }}` and `me`

Ok well, even though I think we should use only one form and remove all other, I add `user_uuid` and `me`

PR #53, opened 2019-11-15
refresh token: do not publish create event when not created

pc-m:
If a refresh token is re-used instead of creating a new one the refresh token
created event is not published


pc-m:
regate

PR #55, opened 2019-11-28
use flush instead of commit to check integrity

fblackburn1:
reason: When there are intermediate commits, we cannot rollback on error
and it can cause weird behaviour

### DIFF
--- a/wazo_auth/database/queries/address.py
+++ b/wazo_auth/database/queries/address.py
@@ -33,7 +33,7 @@ class AddressDAO(BaseDAO):
         address = Address(**kwargs)
         with self.new_session() as s:
             s.add(address)
-            s.commit()
+            s.flush()
             return address.id_
 
     def update(self, address_id, **kwargs):
--- a/wazo_auth/database/queries/external_auth.py
+++ b/wazo_auth/database/queries/external_auth.py
@@ -42,7 +42,7 @@ class ExternalAuthDAO(filters.FilterMixin, PaginatorMixin, BaseDAO):
             external_type = self._find_or_create_type(s, auth_type)
             external_data = ExternalAuthData(data=serialized_data)
             s.add(external_data)
-            s.commit()
+            s.flush()
             user_external_auth = UserExternalAuth(
                 user_uuid=str(user_uuid),
                 external_auth_type_uuid=external_type.uuid,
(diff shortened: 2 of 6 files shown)


fblackburn1:
recheck

fblackburn1:
recheck

fblackburn1:
@fredericlepied Do you have an idea with it fail to fetch requirements?

fredericlepied:
recheck

fredericlepied:
recheck

fblackburn1:
recheck

fblackburn1:
let's go zuul, don't fail the tests and give me a discount on my time for the black friday :smile:

fredericlepied:
recheck

sileht:
I agree that using flush will fix this kind of issue, but the real problem is we have multiple sessions for one http request.

And even with this, the session will be committed at the end of each `with ...` block, making rollback not rollbacking the whole HTTP requests.

`A web application is the easiest case because such an application is already constructed around a single, consistent scope - this is the request.`

https://docs.sqlalchemy.org/en/13/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it

fblackburn1:
> I agree that using flush will fix this kind of issue, but the real problem is we have multiple sessions for one http request.
> 
> And even with this, the session will be committed at the end of each `with ...` block, making rollback not rollbacking the whole HTTP requests.
> 
> `A web application is the easiest case because such an application is already constructed around a single, consistent scope - this is the request.`
> 
> https://docs.sqlalchemy.org/en/13/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it

Totally agree. It's why we had started to scope the request with the session. But in one of your [previous commit](https://github.com/wazo-platform/wazo-auth/commit/93761f69c6e07853392c1acf72956ea8f08eb922) you remove it. So I though you were against this?
Maybe having two methods in the same time is worst than having one of them.

IMO, trying to use `new_session` only once by request isn't worth it, because we cannot guarantee it, and the scoping request with session pattern is the least painful thing

sileht:
> Totally agree. It's why we had started to scope the request with the session. But in one of your [previous commit](https://github.com/wazo-platform/wazo-auth/commit/93761f69c6e07853392c1acf72956ea8f08eb922) you remove it.

Before this change, ton of sessions was open, some was manually flushed, other was never flushed until it reach the flask end request hook. I just wanted to make explicit when transaction really happen.

So I fixup my sqlalchemt session issue, but obviously create new ...

I general flask app have a pre request hook that open the session and a post request hook that close it. so we known exactly when DB transaction happen.

>  So I though you were against this?

I'm not, I would like to have only one sql connection per http request open and to ensure that we must have only one sqlalchemy session open.

> Maybe having two methods in the same time is worst than having one of them.
> 
> IMO, trying to use `new_session` only once by request isn't worth it, because we cannot guarantee it, and the scoping request with session pattern is the least painful thing

PR #57, opened 2019-12-12
Allow tracking HTTP and PostgreSQL requests

lebauce:


fblackburn1:
I don't really understand these changes:
  * Values from `etc/wazo-auth/config.yml` are not used
  * All the magic is in the debian postinst of the package (which is not updated in this branch)
    * For now, all install and upgrade will be unaffected
  * And, even with the postinst fixed, we also need to take care of other private projects :secret:

sduthil:
After talking with François:
* Values from `/etc/wazo-auth/config.yml` are indeed used in the majority of cases
* If pre-configuring the Debian package with debconf, then the values from `/etc/wazo-auth/config.yml` are not used.
Here is a PR to fix the debconf cases: https://github.com/wazo-platform/wazo-auth/pull/65

PR #59, opened 2019-12-24
wazo_default_admin_policy: fix wrong ACL

sduthil:
Why:

* There was an ACL template
"dird.displays.#dird.profiles.#dird.sources.read" in policy
wazo_default_admin_policy #

How:

* If the dird.displays.# is already present in the policy, remove the
wrong ACL
* If the dird.displays.# is not present, replace the wrong ACL template
with dird.displays.#


sduthil:
regate

PR #61, opened 2020-01-06
integration_tests/Makefile: do not use docker pull

fredericlepied:
Depends-On: https://github.com/wazo-platform/xivo-test-helpers/pull/25
Depends-On: https://github.com/wazo-platform/sf-jobs/pull/42

### DIFF
--- a/integration_tests/Makefile
+++ b/integration_tests/Makefile
@@ -1,19 +1,10 @@
-.PHONY: test-setup test test-image test-prerequisite test-db
-
-test-setup: test-prerequisite test-image egg-info test-db build-oauth2-sync
+test-setup: test-image egg-info test-db build-oauth2-sync
 
 build-oauth2-sync:
 	docker build --no-cache -t oauth2-synchronization-service -f assets/external_auth/oauth2_synchronization_service/Dockerfile assets/external_auth/oauth2_synchronization_service
 
 test-db:
 	docker build -f ../contribs/docker/Dockerfile-db -t wazopbx/wazo-auth-db ..
-
-test-prerequisite:
-	docker pull munkyboy/fakesmtp
-	docker pull nickstenning/slapd
-	docker pull rabbitmq
-	docker pull wazopbx/postgres-test
-	docker pull wazopbx/wait
 
 test-image:
 	docker build --no-cache -t wazopbx/wazo-auth ..
@@ -30,3 +21,5 @@ mount:
 
 umount:
 	for f in $$(ls assets/*/docker-compose.yml); do echo "updating $$f" && sed -i 's/- "$${LOCAL_GIT_REPOS}/# - "$${LOCAL_GIT_REPOS}/' $$f; done
+
+.PHONY: test-setup test test-image test-prerequisite test-db
--- a/integration_tests/Makefile
+++ b/integration_tests/Makefile
@@ -16,10 +16,4 @@ egg-info:
 test:
 	pytest -x
 
-mount:
-	for f in $$(ls assets/*/docker-compose.yml); do echo "updating $$f" && sed -i 's/# - "$${LOCAL_GIT_REPOS}/- "$${LOCAL_GIT_REPOS}/' $$f; done
-
-umount:
-	for f in $$(ls assets/*/docker-compose.yml); do echo "updating $$f" && sed -i 's/- "$${LOCAL_GIT_REPOS}/# - "$${LOCAL_GIT_REPOS}/' $$f; done
-
 .PHONY: test-setup test test-image test-db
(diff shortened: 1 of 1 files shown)


fredericlepied:
recheck

integration_tests/Makefile
```diff
@@ -1,23 +1,16 @@
...
-	docker pull munkyboy/fakesmtp
-	docker pull nickstenning/slapd
-	docker pull rabbitmq
-	docker pull wazopbx/postgres-test
-	docker pull wazopbx/wait
-
 test-image:
-	docker build --no-cache -t wazopbx/wazo-auth ..
-	docker build --no-cache -t wazopbx/wazo-auth-tests -f Dockerfile-auth-test ..
+	docker image inspect --format='{{ .Id }}' wazopbx/wazo-auth || docker build --no-cache -t wazopbx/wazo-auth ..
```
sileht: This will make the image not rebuilt when we run tox -e intergration locally?

fredericlepied: yes that's the drawback. We could try to avoid this in sf-jobs in another way.

fblackburn1: Please don't change this behavior. There are enough of corner case with integration tests, we need to be confident that if we do a `make test-setup` we start with a very clean setup and if we have an error this is the code fault, not the environment.

fredericlepied: yes I have found a solution for this. Don't worry ;-)

fblackburn1: Don't see the other commits, please mark the discussion as solved with the decision taken

fredericlepied: The need for cache us solved by avoiding the current PR in https://github.com/wazo-platform/sf-jobs/pull/41

integration_tests/Makefile
```diff
@@ -1,20 +1,11 @@
...
 test-db:
 	docker build -f ../contribs/docker/Dockerfile-db -t wazopbx/wazo-auth-db ..
 
-test-prerequisite:
-	docker pull munkyboy/fakesmtp
-	docker pull nickstenning/slapd
-	docker pull rabbitmq
-	docker pull wazopbx/postgres-test
-	docker pull wazopbx/wait
-
```
fblackburn1: Don't remove docker pull
`docker-compose up` will not pull new image.
Why we need it: https://github.com/wazo-platform/wazo-amid/pull/11#discussion_r363891392

fredericlepied: Why do you say it will not pull a new image as the doc is stating the opposite except if you use --no-recreate: https://docs.docker.com/compose/reference/up/

And this is needed to let docker-comoose take a local fresh build instead of forcing the download from the remote registry. We need it to have docker images built from dependent PRs working.

fblackburn1: the documentation do not state of anything about pulling the image from registry. The only thing it mention is if a CONTAINER is started without the last image on the host, then it will be recreated.
You can try the following scenario
Given new host without any image
Given `docker-compose up` <-- new image will be pulled from registry
When I wait until new image is available on docker registry
When `docker-compose up`
Then nothing is changed, because image already exist on host

The `docker-compose up` command do not pull image from registry if it already exist on host.

I understand problems with zuul and local image, but I not agree with this solution since it can cause other problems for developer environment.

fredericlepied:
recheck

fredericlepied:
recheck

fredericlepied:
recheck

fredericlepied:
recheck

fredericlepied:
recheck

integration_tests/Makefile
fredericlepied: Ok I tried to make it work for both cases by using docker-compose pull automatically except in CI in this: https://github.com/wazo-platform/xivo-test-helpers/pull/25/files#

Let me know if this is ok for you.

fblackburn1: Yes it is a solution that I would like to have time to implement, thanks !! :+1:

integration_tests/test-requirements-for-tox.txt
```diff
@@ -4,7 +4,7 @@ kombu
 mock
 pytest
 openapi-spec-validator
-pyhamcrest
+pyhamcrest!=1.10
```
fblackburn1: The error was fixed by https://github.com/wazo-platform/xivo-test-helpers/pull/24
No need to add version dependency

fblackburn1: fixed: revert this change

fblackburn1:
There are any reason why moving `.PHONY` on bottom of the file?
Is it a Makefile standard or guideline?
Personally I don't care, it's just that all our Makefile have the `PHONY` on top of the file, so I would like to know which standard we should use?

fredericlepied:
> There are any reason why moving `.PHONY` on bottom of the file?
> Is it a Makefile standard or guideline?
> Personally I don't care, it's just that all our Makefile have the `PHONY` on top of the file, so I would like to know which standard we should use?

Yes that's the usual way to have only one .PHONY at the end of the file.

fblackburn1:
@fredericlepied do you miss the mergeit label or you don't want to mergeit for now?

PR #62, opened 2020-01-16
WAZO-1481 alembic: add external auth event ACL to users

sduthil:


alembic/versions/dc2848563b53_add_external_auth_event_to_default_users.py
```diff
@@ -0,0 +1,114 @@
...
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'dc2848563b53'
+down_revision = '4d03433e57fd'
+
+POLICY_NAME = 'wazo_default_user_policy'
+ACL_TEMPLATES = [
+    'events.auth.users.*.external.#'
```
fblackburn1: we don't want to authorize all user to see events for every user. replace the * with the user uuid

PR #69, opened 2020-02-13
WAZO-1638 reduce iteration number of pbkdf2_hmac

safchain:
100 hash generation

Without this PR:
real	0m18.878s
user	0m18.805s
sys	0m0.023s

With it:
real	0m7.876s
user	0m7.816s
sys	0m0.022s

### DIFF
--- a/wazo_auth/services/user.py
+++ b/wazo_auth/services/user.py
@@ -167,7 +167,7 @@ class PasswordEncrypter:
 
     _salt_len = 64
     _hash_algo = 'sha512'
-    _iterations = 250000
+    _iterations = 100000
 
     def encrypt_password(self, password):
         salt = os.urandom(self._salt_len)


sboily:
Nice thank you!

sboily:
Don't work for me. I don't know why but I have 401 with this patch.

pc-m:
There is some more work involved to change the hashing algorithm. I think the following solution would be simple to implement.

1. Add a column `password_hash_algorighm_version` with a default value `0` on the user table
2. Change the default value to `1`
3. Create 2 have 2 functions, one to check passwords against the algorithm version 0 and one for version 1
4. When someone authenticate using its password and it's in version 0 if the login is successful
4.1 Compute the new hash using version 1
4.2 Store the new hash and the hash version

When a new password is created always use the most recent version.

safchain:
@pc-m thanks for the suggestions. Before going further with this I wanted to be sure that we wanted to go that way, meaning changing the authentication algorithm ? I personally would say yes, at least having a upgrade path would worth it.

pc-m:
> @pc-m thanks for the suggestions. Before going further with this I wanted to be sure that we wanted to go that way, meaning changing the authentication algorithm ? I personally would say yes, at least having a upgrade path would worth it.

I think that would certainly help with connections spikes when a service is restarted and tens or hundreds of clients can try to get a new token at the same time.

The 25000 iteration value that we chose was fast enough not to feel too long for the user without taking the responsiveness of wazo-auth itself into account. With the numbers you got I think 10000 is a great compromise.

fblackburn1:
I close this PR, because we don't intend to merge it.
Jira related issue https://wazo-dev.atlassian.net/browse/WAZO-1638

PR #70, opened 2020-02-24
Wazo 1609 request bound sessions

pc-m:
This PR changes the management of database sessions. Sessions are committed at the end of each HTTP session. If an error occurs during the handling of the HTTP request the session will be rolledback.

There are some areas of wazo-auth where requests happen outside the scope of an HTTP request. In those areas I committed and closed the session manually.

* Expired token/session thread
* Bootstrap script
* OAuth2 ws callbacks

### DIFF
--- a/wazo_auth/plugins/external_auth/google/websocket_oauth2.py
+++ b/wazo_auth/plugins/external_auth/google/websocket_oauth2.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2019-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import json
@@ -8,6 +8,7 @@ from threading import Thread
 import websocket
 
 from wazo_auth.exceptions import ExternalAuthAlreadyExists
+from wazo_auth.database.helpers import Session
 from .helpers import get_timestamp_expiration
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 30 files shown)


wazo_auth/plugins/external_auth/google/websocket_oauth2.py
```diff
@@ -50,6 +51,10 @@ def _on_message(self, message):
             msg = json.loads(message)
             self.ws.close()
             self.create_first_token(self.user_uuid, msg.get('code'))
+            try:
+                Session.commit()
+            finally:
+                Session.close()
```
fblackburn1: Can you create a helper to scope these line to avoid duplication
Also, to be clean, we should rollback the session if an error occurs

pc-m:
> nice work  (but not fun to review)
I know sorry about that :(
> 
> We should use flush after EVERY query (before leaving dao) that change the data (create/update/delete) to avoid incomprehensible error at the end of the request
You're right
> 
> I'm not sure about the benefit of using `begin_nested`.
> 
> * for queries: If an error occurs we just want to rollback all transactions for this request. If we commit only a small subset of query, we have the same bad behavior as before? Or we still need to have this logic on our code?
I've added too much begin_nested in the queries. I've removed all of them except 3. The remaining ones are for associations where a duplicate should not generate an error and is just a noop. The commit in a begin_nested only ends that nested block. If another exception happens later during the HTTP request a subsequent rollback will rollback what was in the scoped session.
> * for db fixtures: the test code use the same session than fixture, so we will just revert all test action + the fixture. (and I think that if multiple fixtures, it doesn't care if rollback is called multiple times)
Yes everything that happens within the function that use a fixture is rolled back at the end of the fixture. I did not see a problem with that since the test is done anyway.
> * for base tests: it should be useful if with the same session object we will insert data (before the begin_nested), but it is not the case here. IMO we should be symmetric with the creation of the Session and the removal of this object (see comment about `Session.remove()`). And if we follow that, `begin_nested` is not necessary.

I've pushed more commits to address all you points. Thanks for your review

PR #71, opened 2020-02-27
Wazo 1610 remove cached tenant tree

pc-m:
I tried to come up with a new name for the TenantTree class but could not find one.


fblackburn1:
regate

pc-m:
recheck

pc-m:
regate

PR #80, opened 2020-05-04
test_db_user: avoid search matching autogenerated names

sduthil:
Why:

* When searching "baz", it returns a user with lastname
"rdqkpratcl_baz_gaivare" and username "foo"


sduthil:
Agreed that the test is a bit less readable.

Another method would be to disallow a certain substring during string generation, e.g.
```
@fixtures.db.user(username='foo', email_address='foo@example.com', avoid=['bar', 'baz'])
```

PR #83, opened 2020-05-12
add ACL for dird graphql schema to users

sduthil:
Why:

* Users should be able to get the introspection data from GraphQL, e.g.
to have autocompletion in GraphQL UI.

Jira-Issue: https://wazo-dev.atlassian.net/browse/WAZO-1778


sduthil:
According to https://wazo-dev.atlassian.net/browse/WAZO-1741, the ACL on this query will be removed.

PR #84, opened 2020-05-19
test_tokens: avoid autogenerating attributes when searching

sduthil:
Why:

* Tests sometimes fail because the autogenerated client_id matches "baz"


pc-m:
regate

PR #103, opened 2020-09-03
Pull request for WAZO-1958-group-for-all-users

sduthil:
tenants: create wazo-all-users group with new tenants
tenants: add wazo-all-users groups to existing tenants
users: add new users into wazo-all-users group
users: add existing users to wazo-all-users group
wazo-all-users migration: optimize queries
wazo-all-users: rename already existing groups

### DIFF
--- a/alembic/versions/16f98957ec7e_add_wazo_all_users_group.py
+++ b/alembic/versions/16f98957ec7e_add_wazo_all_users_group.py
@@ -1,0 +1,52 @@
+"""add wazo-all-users group
+
+Revision ID: 16f98957ec7e
+Revises: 91e8de642bee
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '16f98957ec7e'
+down_revision = '91e8de642bee'
+
+from alembic import op
+import sqlalchemy as sa
+
+group_table = sa.sql.table(
+    'auth_group', sa.Column('uuid'), sa.Column('tenant_uuid'), sa.Column('name')
+)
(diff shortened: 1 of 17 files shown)


alembic/versions/16f98957ec7e_add_wazo_all_users_group.py
```diff
@@ -0,0 +1,105 @@
...
+    group_name_query = sa.sql.select(
+        [sa.sql.functions.concat('wazo-all-users-tenant-', tenant_table.c.uuid)]
+    ).select_from(tenant_table)
+    group_uuid_query = sa.sql.select([group_table.c.uuid]).where(
+        group_table.c.name.in_(group_name_query)
+    )
+    update_query = (
+        group_table.update()
+        .values(
+            {group_table.c.name: sa.sql.functions.concat(group_table.c.name, '-old')}
```
fblackburn1, line 42: FYI we already [write this kind of logic](https://github.com/wazo-platform/xivo-manage-db/blob/master/alembic/versions/43995f4ac823_fix_invalid_context_name.py#L40) to avoid to have an error if someone already has `wazo-all-users-tenant-<uuid>-old` which is probably *impossible*

This is a consequence to do not fix our technical debt with global unique name

fblackburn1 (dismissed):
black format on sqlalchemy request is just unreadable... 
To be able to review alembic script I needed to format it manually :frowning_face: 

I approve this PR  because there is no blocking issue, and Monday we will start to merge pjsip and I don't want to debug two big behavior changes at the same time and/or delay this merge

fblackburn1 (approved):

PR #106, opened 2020-09-09
for WAZO-1960: check multiple scopes

bloom1:
Depends-On: https://github.com/wazo-platform/wazo-auth-client/pull/15

### DIFF
--- a/integration_tests/suite/test_http_interface.py
+++ b/integration_tests/suite/test_http_interface.py
@@ -198,8 +198,12 @@ class TestCore(WazoAuthTestCase):
 
         assert_that(self._is_valid(token, tenant=self.top_tenant_uuid), is_(True))
 
-        with self.client_in_subtenant() as (_, __, sub_tenant):
+        with self.client_in_subtenant() as (sub_client, __, sub_tenant):
             assert_that(self._is_valid(token, tenant=sub_tenant['uuid']), is_(True))
+            assert_that(
+                self._is_valid(sub_client._token_id, tenant=self.top_tenant_uuid),
+                is_(False),
+            )
 
     def test_that_unauthorized_acls_on_GET_return_403(self):
         token = self._post_token('foo', 'bar')['token']
--- a/integration_tests/suite/test_http_interface.py
+++ b/integration_tests/suite/test_http_interface.py
@@ -240,6 +240,16 @@ class TestCore(WazoAuthTestCase):
 
         self._get_token(token, acls='foo')  # no exception
 
+    def test_that_expired_tokens_on_scope_check_returns_404(self):
+        token = self._post_token('foo', 'bar', expiration=1)['token']
+
+        time.sleep(2)
+
+        assert_that(
+            calling(self._check_scopes).with_args(token, ['foo']),
+            raises(requests.HTTPError, pattern='404'),
+        )
+
     def test_that_scope_check_with_an_invalid_token_returns_404(self):
         assert_that(
             calling(self._check_scopes).with_args('abcdef', []),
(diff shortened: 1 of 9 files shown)


integration_tests/suite/test_http_interface.py
```diff
@@ -240,6 +241,67 @@ def test_that_authorized_acls_on_GET_return_200(self, policy):
...
+            not_(raises(Exception)),
+        )
+
+        with self.client_in_subtenant() as (_, __, sub_tenant):
+            assert_that(
+                calling(self._check_scopes).with_args(
+                    token, ['foo'], tenant=sub_tenant['uuid']
+                ),
+                not_(raises(Exception)),
+            )
```
sduthil, line 280: A test is missing: if client is in a subtenant, checking against a top tenant should return 403.

integration_tests/suite/test_http_interface.py
```diff
@@ -240,6 +241,67 @@ def test_that_authorized_acls_on_GET_return_200(self, policy):
...
+            self._check_scopes(token, ['confd', 'foo', 'bar']),
+            has_entries(confd=False, foo=False, bar=False),
+        )
+
+    def test_that_unauthorized_tenants_on_scope_check_return_403(self):
+        token = self._post_token('foo', 'bar')['token']
+
+        assert_that(
+            calling(self._check_scopes).with_args(
+                token, ['foo'], tenant='55ee61f3-c4a5-427c-9f40-9d5c33466240'
```
sduthil, line 262: Magic number here: could you replace with a constant like `UNKNOWN_TENANT`?

wazo_auth/plugins/http/tokens/http.py
```diff
@@ -20,6 +20,25 @@ def __init__(self, token_service, user_service, authentication_service):
         self._authentication_service = authentication_service
 
 
+class TenantCheckMixin:
```
sduthil, line 23: Nice deduplication!
However, I'd prefer this logic to be in the service rather than the HTTP interface.
Also, I'd prefer this to be a function rather than a mixin method. The Mixin class uses `self._user_services`, but this attribute is defined in the other class.

wazo_auth/services/token.py
```diff
@@ -166,6 +166,26 @@ def get(self, token_uuid, required_acl):
...
+
+        if token.is_expired():
+            raise UnknownTokenException()
+
+        scope_statuses = dict()
+
+        for scope in scopes:
+            if scope in scope_statuses:
+                continue
+            scope_statuses[scope] = token.matches_required_acl(scope)
```
sduthil, line 185: More Pythonic:
```python
scope_statuses = {scope: token.matches_required_acl(scope) for scope in set(scopes)}
```

wazo_auth/services/token.py
```diff
@@ -166,6 +166,26 @@ def get(self, token_uuid, required_acl):
...
+    def check_scopes(self, token_uuid, scopes):
+        token_data = self._dao.token.get(token_uuid)
+        if not token_data:
+            raise UnknownTokenException()
+
+        id_ = token_data.pop('uuid')
+        token = Token(id_, **token_data)
+
+        if token.is_expired():
+            raise UnknownTokenException()
```
sduthil, line 178: Nice catch :) Wouldn't have thought of that one. Too bad we can't test it...

wazo_auth/plugins/http/tokens/api.yml
```diff
@@ -116,7 +116,7 @@ paths:
       parameters:
         - $ref: '#/parameters/token'
         - $ref: '#/parameters/scope'
-        - $ref: '#/parameters/tenantuuid'
+        - $ref: '#/parameters/tenant'
```
sduthil, line 119: Thank you for fixing this!

sduthil (changes requested):

sduthil (approved):

bloom1:
regate

bloom1:
recheck

bloom1:
recheck

PR #123, opened 2020-11-13
groups: add searchable columns to openapi doc

sduthil:


wazo_auth/plugins/http/groups/api.yml
```diff
@@ -155,6 +159,29 @@ parameters:
...
+    name: name
+    in: query
+    type: string
+    description: The name of the group
+    required: false
+  search_user_uuid:
+    name: user_uuid
+    in: query
+    type: string
+    description: The UUID of the user
```
chartrandf, line 178: should we add `required: false` too ?

chartrandf (dismissed):

wazo_auth/plugins/http/groups/api.yml
sduthil: yes, done. Thank you!

PR #133, opened 2021-01-13
add amid access to admin user

fblackburn1:
reason: since amid is restricted to master tenant only, it doesn't
impact to have this access on every admin


pc-m (approved):

pc-m:
regate

PR #136, opened 2021-03-16
WAZO-2292: add confd e-mails ACL to admin policy

bloom1:


alembic/versions/973a08801fde_add_emails_configuration_acl.py
```diff
@@ -0,0 +1,109 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '973a08801fde'
+down_revision = '67013e93544f'
+
+POLICY_NAME = 'wazo_default_admin_policy'
+ACL = ['confd.emails.#']
```
fblackburn1, line 16: admin already have `confd.#` access

fblackburn1 (changes requested):

PR #159, opened 2021-05-26
add /config API, allowing change of debug setting

sduthil:
Depends-On: https://github.com/wazo-platform/wazo-auth-client/pull/21


sduthil:
recheck

wazo_auth/plugins/http/config/http.py
```diff
@@ -0,0 +1,26 @@
...
+from .schemas import config_patch_schema
+
+
+class ConfigResource(AuthResource):
+    def __init__(self, config_service):
+        self._config_service = config_service
+
+    @required_acl('auth.config.read')
+    def get(self):
+        return self._config_service.get_config(), 200
```
fblackburn1, line 18: Should be restricted to master tenant
database credential and other sensitive information can be retrieved

fblackburn1 (commented):
Just to be sure `top` vs `master` (in other services) I guess it's because wazo-auth doesn't have `master` terminology?

fblackburn1 (approved):

sduthil:
> Just to be sure `top` vs `master` (in other services) I guess it's because wazo-auth doesn't have `master` terminology?

Yes, exactly.

PR #162, opened 2021-06-17
policy: fix pagination with limit

fblackburn1:
why: refactor from 21.07 remove group_by clause. Now this request return
duplicate rows.

### DIFF
--- a/integration_tests/suite/database/test_db_policy.py
+++ b/integration_tests/suite/database/test_db_policy.py
@@ -262,7 +262,9 @@ class TestPolicyDAO(base.DAOTestCase):
     @fixtures.db.user()
     @fixtures.db.policy(name='name1', acl=['1', '2'])
     @fixtures.db.policy(name='name2', acl=['3', '4'])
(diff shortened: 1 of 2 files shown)


integration_tests/suite/database/test_db_policy.py
```diff
@@ -258,6 +258,23 @@ def test_list_by_read_only(self, policy_1, policy_2):
             ),
         )
 
+    @fixtures.db.user()
+    @fixtures.db.user()
+    @fixtures.db.policy(name='name1', acl=['1', '2'])
+    @fixtures.db.policy(name='name2', acl=['3', '4'])
+    def test_list_group_by_policy(self, user_1, user_2, policy_1, policy_2):
+        self._user_dao.add_policy(user_1, policy_1)
+        self._user_dao.add_policy(user_2, policy_1)
```
sduthil, line 267: Why `policy_1`? What's the use of `policy_2` in this case?

sduthil: Nevermind, I understand the test, I'll rename it.

sduthil (approved):

PR #169, opened 2021-07-29
Pull request for WAZO-2452-boostrap-container-ready

fblackburn1:


sduthil:
Is `wazo-auth-bootstrap complete` still useful after this?
Is `wazo-auth-bootstrap` even useful after this?

sduthil (approved):

sduthil:
Since we're allowing plain credentials in a config file now, what do you think about checking the config files permissions, to ensure they are not readable by other users on the system? Just throwing the idea out there, I'm not sure it's a good idea...

fblackburn1:
> Is `wazo-auth-bootstrap complete` still useful after this?
> Is `wazo-auth-bootstrap` even useful after this?

Still used by debian postinst
The `complete` add wrap-up around wazo-auth-cli (automate password generation and add file in HOME).
If you provide enable automatic bootstrap via config file, but without providing password, an error will be raised

I did not want to change all wazo-auth bootstrap (it was already out-scope). But yes, I think we should remove it to only use config file (in the future).

> Since we're allowing plain credentials in a config file now, what do you think about checking the config files permissions, to ensure they are not readable by other users on the system?

I think we have the same issue with all our config file (database/rabbitmq credentials, wazo-auth service user without key_file, etc..)
In practice, this config file is only used in docker environment, which is less "risky". But I agree that situation is not ideal and may be improved. However, I don't consider that to implement this feature is more risky than before

PR #173, opened 2021-09-21
policies: restrict negative access removal

sduthil:
Why:

* A user could escalate privileges by removing the negative accesses
applying to itself

Depends-On: https://github.com/wazo-platform/xivo-lib-python/pull/154

### DIFF
--- a/integration_tests/suite/helpers/fixtures/http.py
+++ b/integration_tests/suite/helpers/fixtures/http.py
@@ -215,6 +215,7 @@ def policy(**policy_args):
                     result = decorated(self, *args, **kwargs)
             finally:
                 try:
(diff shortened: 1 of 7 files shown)


sduthil:
recheck

sduthil:
recheck

rbienvenu0:
recheck

pc-m:
I tried creating a Policy with `!foo.*.bar`, assigned the policy to user a who has the necessary permission to modify policies. Tried to remove the `!.foo.*.bar`  from the policy and got a nice error. The I used a token from wazo-auth-cli and was able to remove the policy successfully. LGTM

pc-m (approved):

pc-m:
regate

PR #174, opened 2021-10-12
Pull request for WAZO-2559-default-user-policy

fblackburn1:


wazo_auth/services/token.py
```diff
@@ -210,7 +211,12 @@ def check_scopes(self, token_uuid, scopes):
         return token, scope_statuses
 
     def _get_acl(self, backend_name):
-        policy_name = self._backend_policies.get(backend_name)
+        # 21.14: deprecated
+        policy_name = self._deprecated_backend_policies.get(backend_name)
```
sduthil, line 215: Why does the deprecated option take precedence? Shouldn't it be reversed?

wazo_auth/config.py
```diff
@@ -67,7 +67,7 @@
         'microsoft': True,
         'mobile': True,
     },
-    'backend_policies': {},
+    'backend_policies': {},  # 21.14: Deprecated
```
sduthil, line 70:
```suggestion
    'backend_policies': {},  # since 21.14: Deprecated
```

wazo_auth/services/token.py
sduthil: You're right, thanks for the precision.

sduthil (approved):

PR #180, opened 2021-11-23
Wazo-658: Change 500 error message when deleting tenant with children to a user-freindly message

nissou91:
Some integration tests failed on my machine, but I think that the issue in the ticket (WAZO-658) is fixed.

### DIFF
--- a/integration_tests/suite/test_db_tenant.py
+++ b/integration_tests/suite/test_db_tenant.py
@@ -4,7 +4,6 @@
 import pytest
 
 from uuid import uuid4
-
 from hamcrest import (
     assert_that,
     calling,
@@ -23,6 +22,9 @@ from wazo_auth import exceptions
 from wazo_auth.database import models
 
 from .helpers import fixtures, base, constants
+from .helpers.base import (
+    assert_no_error,
+)
 
 TENANT_UUID = '00000000-0000-4000-9000-000000000000'
 USER_UUID = '00000000-0000-4000-9000-111111111111'
@@ -224,6 +226,18 @@ class TestTenantDAO(base.DAOTestCase):
             calling(self._tenant_dao.delete).with_args(tenant_uuid),
             raises(exceptions.UnknownTenantException),
         )
+
+    @fixtures.db.tenant()
+    def test_delete_tenant_with_children(self, tenant_uuid):
+
+        a_uuid = self._create_tenant(name='a', parent_uuid=tenant_uuid)
+
+        assert_that(
+            calling(self._tenant_dao.delete).with_args(tenant_uuid),
+            raises(exceptions.UnauthorizedTenantwithChildrenDelete),
+        )
+
+        assert_no_error(self._tenant_dao.delete, a_uuid)
 
     @fixtures.db.tenant(uuid=TENANT_UUID)
     @fixtures.db.address(id_=ADDRESS_ID, tenant_uuid=TENANT_UUID)
--- a/integration_tests/suite/test_tenants.py
+++ b/integration_tests/suite/test_tenants.py
@@ -229,6 +229,15 @@ class TestTenants(base.APIIntegrationTest):
         assert_no_error(self.client.tenants.delete, tenant['uuid'])
         assert_http_error(404, self.client.tenants.delete, tenant['uuid'])
 
+    @fixtures.http.tenant()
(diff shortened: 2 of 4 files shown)


wazo_auth/database/queries/tenant.py
```diff
@@ -104,7 +104,19 @@ def delete(self, uuid):
         tenant = self.session.query(Tenant).get(uuid)
         if not tenant:
             raise exceptions.UnknownTenantException(uuid)
-        self.session.delete(tenant)
+        
+        children_list = self.session.query(Tenant).filter(Tenant.parent_uuid == uuid)
+        
+        if children_list:
```
pc-m, line 110: `children_list` is a query object not the result of the query. It's going to enter the `if` even if there's no parent_uuid. This causes other tests to fail.

pc-m (changes requested):

integration_tests/suite/test_db_tenant.py
```diff
@@ -225,6 +228,18 @@ def test_delete(self, tenant_uuid):
             raises(exceptions.UnknownTenantException),
         )
 
+    @fixtures.db.tenant()
+    def test_delete_tenant_with_children(self, tenant_uuid):
+
+        a_uuid = self._create_tenant(name='a', parent_uuid=tenant_uuid)
+        self._tenant_dao.delete(tenant_uuid)
```
pc-m, line 235: You should delete the `self._tenant_dao.delete(tenant_uuid)` line because it deletes the tenant before the assertion that tries to delete it

integration_tests/suite/test_db_tenant.py
```diff
@@ -225,6 +228,18 @@ def test_delete(self, tenant_uuid):
...
+
+        assert_that(
```
pc-m, line 237: The assertion should be in a try finally to make sure that tenant `a` gets deleted at the end of the test

integration_tests/suite/test_db_tenant.py
```diff
@@ -225,6 +228,18 @@ def test_delete(self, tenant_uuid):
...
+            calling(self._tenant_dao.delete).with_args(tenant_uuid),
+            raises(exceptions.UnauthorizedTenantwithChildrenDelete),
+        )
+        assert_no_error(self._tenant_dao.delete, a_uuid['uuid'])
```
pc-m, line 241: Should be `a_uuid` instead of `a_uuid['uuid']`

integration_tests/suite/test_db_tenant.py
```diff
@@ -4,6 +4,9 @@
 import pytest
 
 from uuid import uuid4
+from .helpers.base import (
```
pc-m, line 7: Could you move the import lower, after the `from .helper import` line.
We usually use the following order.
1. standard library
2. other library
3. imports from this project
4. relative imports

integration_tests/suite/test_tenants.py
```diff
@@ -229,6 +229,11 @@ def test_delete(self, tenant):
         assert_no_error(self.client.tenants.delete, tenant['uuid'])
         assert_http_error(404, self.client.tenants.delete, tenant['uuid'])
 
+    @fixtures.http.tenant()
+    def test_delete_tenant_with_children(self, tenant):
+        with self.client_in_subtenant() as (client, user, sub_tenant):
```
pc-m, line 234: You will need to specify the parent_uuid when calling client_in_subtenant to make sub_tenant a child of tenant.
`with self.client_in_subtenant(parent_uuid=tenant['uuid']) as (client, user, sub_tenant):`

integration_tests/suite/test_tenants.py
```diff
@@ -229,6 +229,11 @@ def test_delete(self, tenant):
...
+            assert_http_error(400, client.tenants.delete, tenant['uuid'])
```
pc-m, line 235: The client returned by `client_in_subtenant` is scoped to the new subtenant and cannot delete a resource "above" sub_tenant. You should use `self.client.tenants.delete` instead.

pc-m (changes requested):
Hi Aniss,

I've added inline comments in the PR.
You should also rename the PR for something more descriptive.
You could also use `git rebase -i master` to update your commit message or group some of them together.

pc-m (approved):

PR #186, opened 2021-12-21
Wazo-607: Return 400 error message when tenant address is empty (at tenant creation and update)

nissou91:


fblackburn1 (changes requested):
Ticket mentions error 500 when empty string
This solution is kind a feature. i.e. accept empty string, which we don't want in our API
Because it can lead to confusion for one meaning with two different values (`null` vs `' '`) and can make migration script more complicate

To solve this, we should return a standard 400 error. ex:
```json
{
  "message": "Length must be between 1 and 128.",
  "error_id": "invalid-data",
  "details": {
    "address": { 
        "line_1": {
            "constraint_id": "length",
            "constraint": {
              "min": 1,
              "max": 128
            },
           "message": "Length must be between 1 and 128."
          }
    }
  },
  "timestamp": 1641569822.5895898,
  "resource": "tenants"
}
```
This is only a example, I'm not sure how the code will generate embedded field error.

Note: Avoid merge master branch into your branch, prefer rebasing your branch

PR #193, opened 2022-01-06
Wazo-2638: Migration of Wazo Auth from Marshmallow 3.0.0b14 to 3.10.0.

nissou91:
Depends-On: https://github.com/wazo-platform/xivo-lib-python/pull/93


fblackburn1:
recheck

PR #196, opened 2022-01-18
Wazo-607: return 400 http error message when tenant's address is empty (at creation and update)

nissou91:


wazo_auth/plugins/http/tenants/tests/test_tenants.py
```diff
@@ -68,15 +88,56 @@ def test_that_validated_args_are_passed_to_the_service(self, TenantDetector):
...
-                country=None,
-            ),
+            address={
+                'line_1': 'xxx',
+                'line_2': 'xxx',
+                'city': 'Montreal',
+                'state': 'Quebec',
+                'country': 'Canada',
+                'zip_code': 'XXX',
+            },
```
fblackburn1, line 98: These changes are not necessary

wazo_auth/exceptions.py
```diff
@@ -177,7 +177,11 @@ def from_errors(cls, errors):
             for info in infos:
                 if isinstance(info, (str, bytes)):
                     info = {'message': info}
-                return cls(info['message'], {field: info})
+                if 'message' in info:
+                    return cls(info['message'], {field: info})
+                else:
+                    for sub_field, sub_info in info.items():
+                        return cls(sub_info[0]['message'], {sub_field: sub_info[0]})
```
fblackburn1, line 184: Should handle case when sub_info is empty
Should avoid to use two different pattern for the same thing (using [0] vs loop and return first item)

fblackburn1 (changes requested):
* commit message should be reword
* parent field should be in the error

fblackburn1 (approved):
fixed - see commits
About the rebase of your commit - I removed unnecessary changes to a test and reword message

PR #197, opened 2022-01-20
WP-494: active directory compatiblity

bloom1:
Essentially, it is already compatible. The important part is that Microsoft uses a particular DN format.
It looks like: `CN=Firstname Lastname,DN=Users,DC=domain,DC=ext` instead of the usual `CN=Firstname Lastname,OU=People,DC=domain,DC=ext`
However, I added some tests to support some cases I stumbled upon.

### DIFF
--- a/integration_tests/assets/docker-compose.ldap.override.yml
+++ b/integration_tests/assets/docker-compose.ldap.override.yml
@@ -8,10 +8,6 @@ services:
     environment:
       TARGETS: "slapd:389 postgres:5432 auth:9497"
 
-  postgres:
-    volumes:
-      - "./postgres_data/populate.ldap.sql:/pg-init-db/populate.sql"
-
   auth:
     volumes:
       - "./etc/wazo-auth/conf.d/asset.ldap.yml:/etc/wazo-auth/conf.d/asset.ldap.yml"
--- a/integration_tests/assets/docker-compose.ldap_anonymous.override.yml
+++ b/integration_tests/assets/docker-compose.ldap_anonymous.override.yml
@@ -8,10 +8,6 @@ services:
     environment:
       TARGETS: "slapd:389 postgres:5432 auth:9497"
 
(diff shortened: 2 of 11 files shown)


pc-m (commented):
The "link" between the ldap user and the Wazo should user should be done using the wazo-auth user instead of using the `find_by` from xivo-dao.
Is it planned for another story?

bloom1:
> 

No it is not planned in another story. Good point, I will get rid of that monstrosity in this ticket directly.

Thank you!

pc-m (changes requested):
`get_metadata` should be modified to find the purpose of the user and generate the metadata from the purpose plugin associated to that purpose.

The change should be pretty simple if you inspire yourself from the wazo_user plugin.

bloom1:
> 

Hello @pc-m, I have an issue with the metadata plugin:

Given the LDAP is configured for a login with a DN of `CN=Alice Wonderland,OU=people,OU=quebec,DC=wazo,DC=io` then the username is `Alice Wonderland`. It means that that LDAP username MUST match the one in the wazo-auth database. Do we really want an association of the username, given we already have the email association?

bloom1:
> > `get_metadata` should be modified to find the purpose of the user and generate the metadata from the purpose plugin associated to that purpose.
> > The change should be pretty simple if you inspire yourself from the wazo_user plugin.
> 
> Hello @pc-m, I have an issue with the metadata plugin:
> 
> Given the LDAP is configured for a login with a DN of `CN=Alice Wonderland,OU=people,OU=quebec,DC=wazo,DC=io` then the username is `Alice Wonderland`. It means that that LDAP username MUST match the one in the wazo-auth database. Do we really want an association of the username, given we already have the email association?

Forget what I said, I am able ot add an email address to the `args` that is passed around and it works wonderfully :smile:

pc-m (approved):

PR #198, opened 2022-01-21
LDAP login API spec

bloom1:


wazo_auth/plugins/http/tokens/api.yml
```diff
@@ -37,6 +37,11 @@ paths:
             backend:
               type: string
               default: wazo_user
+            tenant_id:
+              type: string
+              description: |
+                The `tenant_id` is the identifier in which the token should be created. This is useful only in cases
+                when the backend used is external, i.e not `wazo_user`.
```
fblackburn1, line 44: Can you specify which identifier you talk about? (UUID or slug or both?)

wazo_auth/plugins/http/tokens/api.yml
```diff
@@ -37,6 +37,11 @@ paths:
             backend:
               type: string
               default: wazo_user
+            tenant_id:
```
pc-m, line 40: I don't think we need to add this parameter. The tenant can be found based on the username being provided. I'm also unsure how the users could know that ID without relaying on another API that would "leak" all the tenants of a givens stack to the users.

bloom1: We discussed this on Mattermost - it is possible for an LDAP user to use any kind of username (`Firstname Lastname` even). There needs to be a match made by the backend to find the email address in that case... The goal is to have one LDAP source per tenant. We do not want to have to try the username-password combination on every LDAP server to find the email address because it leaks the username-password combination of the user to all the external LDAP servers. 

The `tenant_id` field is mainly supposed to be used with the slug and given by administrators to the tenant's users.

pc-m (dismissed):

bloom1:
Why the hell did Zuul merge this?! There was no `mergeit` label!

sduthil:
This PR was merged because the commits in this PR were also part of https://github.com/wazo-platform/wazo-auth/pull/200

PR #199, opened 2022-01-31
WP-497 spec ldap config

bloom1:
Do not merge, WAZO-2653 https://github.com/wazo-platform/wazo-auth/pull/200 is based on this PR and this will be merged with it :smile:

### DIFF
--- a/wazo_auth/plugins/http/ldap_config/api.yml
+++ b/wazo_auth/plugins/http/ldap_config/api.yml
@@ -1,0 +1,153 @@
+paths:
+  /backends/ldap:
+    get:
+      security:
+        - wazo_auth_token: []
+      produces:
+        - application/json
+      summary: Get current tenant's LDAP backend configuration.
+      description: '**Required ACL:** `auth.backends.ldap.read`'
+      operationId: getLDAPBackendConfig
+      tags:
+        - backends
+      parameters:
+        - $ref: '#/parameters/tenantuuid'
+      responses:
+        '200':
+          description: The LDAP backend configuration
+          schema:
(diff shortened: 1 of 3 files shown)


wazo_auth/plugins/http/ldap_config/api.yml
```diff
@@ -0,0 +1,82 @@
+paths:
+  /backends/ldap_user:
```
fblackburn1, line 2: Preference for `/backends/ldap`

wazo_auth/plugins/http/ldap_config/api.yml
```diff
@@ -0,0 +1,82 @@
...
+      tags:
+        - backends
+      parameters:
+        - name: body
+          in: body
+          description: The LDAP backend configuration
+          required: true
+          schema:
+            $ref: '#/definitions/LDAPBackendConfig'
+        - $ref: '#/parameters/tenantuuid'
```
fblackburn1, line 47: Should be before the body to follow other API style

wazo_auth/plugins/http/ldap_config/api.yml
bloom1: Nice, indeed it is better. Thank you :smiley:

wazo_auth/plugins/http/ldap_config/api.yml
bloom1: I agree, thank you :smile:

fblackburn1 (approved):

bloom1:
Merged in PR #200

PR #203, opened 2022-02-08
Pull request for WAZO-2678-unique-username-email

fblackburn1:
No migration has been done: The use case for having username == email from
another user is more an exception than normal behvior
In this case, on the next user update, a 409 will be prompted


pc-m (dismissed):

fblackburn1:
recheck

pc-m (approved):

pc-m:
recheck

PR #204, opened 2022-02-10
WAZO-2497 bus inject tenant_uuid

ja0x:


wazo_auth/token.py
```diff
@@ -189,5 +189,6 @@ def _publish_event(self, tokens, sessions, event_class):
                 logger.warning(
                     'session without token associated: {}'.format(session['uuid'])
                 )
-
-            self._bus_publisher.publish(event_class(**event_args))
+            tenant_uuid = event_args.get('tenant_uuid', None)
+            headers = {'tenant_uuid': tenant_uuid} if tenant_uuid else None
```
fblackburn1, line 193: Don't use `get` when `tenant_uuid` is always defined in `event_args`

```suggestion
            tenant_uuid = event_args['tenant_uuid']
            headers = {'tenant_uuid': tenant_uuid} if tenant_uuid else None
```
or
```suggestion
            headers = None
            if event_args['tenant_uuid']:
                headers = {'tenant_uuid': event_args['tenant_uuid']}
```

integration_tests/suite/test_external.py
```diff
@@ -196,11 +209,19 @@ def oauth2_is_done():
         until.assert_(oauth2_is_done, timeout=5, interval=0.25)
 
         def bus_received_msg():
+            msg = msg_accumulator.accumulate(with_headers=True)
+            print(msg)
```
fblackburn1, line 213: Not the first time you forget `print`. You can add pre-commit hook to detect them. [Example](https://github.com/wazo-platform/xivo-tools/blob/master/git-hooks/local-docker-volume-check)
```suggestion
```

integration_tests/suite/test_external.py
```diff
@@ -196,11 +209,19 @@ def oauth2_is_done():
         until.assert_(oauth2_is_done, timeout=5, interval=0.25)
 
         def bus_received_msg():
+            msg = msg_accumulator.accumulate(with_headers=True)
```
fblackburn1, line 212: I guess you move this line outside of the `assert_that` only to print result?

fblackburn1 (changes requested):
every work as expected :+1:

wazo_auth/services/external_auth.py
```diff
@@ -148,3 +155,10 @@ def update(self, user_uuid, auth_type, data):
 
     def update_config(self, auth_type, data, tenant_uuid):
         return self._dao.external_auth.update_config(auth_type, data, tenant_uuid)
+
+    def _get_tenant_uuid_from_user(self, user_uuid):
+        try:
+            user = self._dao.user.list_(uuid=user_uuid, limit=1)[0]
+            return user['tenant_uuid']
```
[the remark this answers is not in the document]
ja0x: IndexError is to catch out of bound errors, meaning returning None if no user are in the list returned by the query.  I do agree with the previous statement, the exception should be raised letting the program know that the requested user does not exist rather than silently returning none, will be changed

integration_tests/suite/test_external.py
ja0x: :see_no_evil:
Good idea with the hook

wazo_auth/token.py
```diff
@@ -174,6 +174,7 @@ def _tokens_notice(self):
         self._publish_event(tokens, sessions, SessionExpireSoonEvent)
 
     def _publish_event(self, tokens, sessions, event_class):
+        headers = {}
```
fblackburn1, line 177: I don't think it's a good practice to reuse same dict for different purpose/value.
Should be defined just before the `if event_args['tenant_uuid']`

fblackburn1 (approved):

PR #206, opened 2022-02-14
add ldap_config in config.py

fblackburn1:
why: be explicit about every plugin embedded by wazo-auth


bloom1:
Good point

However, I was about to create a PR to enable both the LDAP plugins by default :smiley: I talked with @sboily about it and there is no reason for the plugin to be disabled by default now.

fblackburn1:
> Good point
> 
> However, I was about to create a PR to enable both the LDAP plugins by default smiley I talked with @sboily about it and there is no reason for the plugin to be disabled by default now.

Ok but don't forget to disable it by default where it should not be enabled by default ... :sweat_smile:

fblackburn1:
Superseded by https://github.com/wazo-platform/wazo-auth/pull/207

PR #207, opened 2022-02-14
WP-543 ldap: enable backend/http plugins by default

bloom1:
Why: no reason not to anymore.


fblackburn1 (commented):
Since we enable ldap by default, can you check to remove test config file `asset.ldap.yml`.

fblackburn1 (approved):

bloom1:
not sure that the fail is a real fail

bloom1:
recheck

PR #208, opened 2022-02-14
ldap_config: remove POST in API

bloom1:
Why: it does the same thing as the PUT and is not useful in that case. The GET
will return a schema with all fields set to null instead of returning a 404
when there is no config in the database.


bloom1:
recheck

fblackburn1 (approved):

PR #212, opened 2022-02-23
WAZO-2699: LDAP refresh token

bloom1:
Technical details: I replace the `login` provided by the user when creating a refresh token by the one that matched during password verification in the Wazo user database.


pc-m (approved):

bloom1:
regate

PR #219, opened 2022-04-05
Add domain_names field to wazo-auth

nissou91:
**Reason:**

As a developer, I would like to be able to set a list of domain names for my tenant.

**Constraints:**

* Must be unique,
* Must contain valid domain names,
* Must support all CRUD operations,
* Must support search for a specific domain_name

### DIFF
--- a/integration_tests/suite/test_db_tenant.py
+++ b/integration_tests/suite/test_db_tenant.py
@@ -403,3 +403,25 @@ class TestTenantDAO(base.DAOTestCase):
 
         result = self._tenant_dao._tenant_query(TENANT_UUID_2)
         assert_that(result[0], has_properties(name='2-updated'))
+
+    @fixtures.db.tenant(name='1', uuid=TENANT_UUID_1, domain_names=VALID_DOMAIN_NAMES_1)
+    def test_update_tenant_domain_names(self, *_):
+
+        self._tenant_dao.update(
+            TENANT_UUID_1, name='1-updated', domain_names=['wazo.io']
+        )
+
+        self._assert_tenant_matches(
+            TENANT_UUID_1, '1-updated', domain_names=['wazo.io']
+        )
+
+    @fixtures.db.tenant(name='1', uuid=TENANT_UUID_1, domain_names=VALID_DOMAIN_NAMES_1)
+    @fixtures.db.tenant(name='2', uuid=TENANT_UUID_2, domain_names=VALID_DOMAIN_NAMES_2)
+    def test_update_tenant_with_duplicate_domain_names_raises_409(self, *_):
+
+        assert_that(
+            calling(self._tenant_dao.update).with_args(
+                tenant_uuid=TENANT_UUID_2, domain_names=['wazo.io']
+            ),
+            raises(exceptions.DomainAlreadyExistException),
+        )
--- a/integration_tests/suite/test_db_tenant.py
+++ b/integration_tests/suite/test_db_tenant.py
@@ -127,9 +127,9 @@ class TestTenantDAO(base.DAOTestCase):
             ),
         )
 
-    @fixtures.db.tenant(name='a', slug='zzz')
-    @fixtures.db.tenant(name='b', slug='yyy')
-    @fixtures.db.tenant(name='c', slug='xxx')
+    @fixtures.db.tenant(name='a', slug='zzz', domain_names=VALID_DOMAIN_NAMES_1)
+    @fixtures.db.tenant(name='b', slug='yyy', domain_names=VALID_DOMAIN_NAMES_2)
+    @fixtures.db.tenant(name='c', slug='xxx', domain_names=VALID_DOMAIN_NAMES_3)
     def test_count(self, *tenants):
         top_tenant_uuid = self._top_tenant_uuid()
         visible_tenants = tenants + (top_tenant_uuid,)
@@ -150,9 +150,23 @@ class TestTenantDAO(base.DAOTestCase):
         result = self._tenant_dao.count(visible_tenants, name='b')
         assert_that(result, equal_to(1))
 
-    @fixtures.db.tenant(name='baz a')
-    @fixtures.db.tenant(name='bar b')
-    @fixtures.db.tenant(name='foo c')
+        result = self._tenant_dao.count(
+            visible_tenants, domain_name='stack.dev.wazo.io'
+        )
+        assert_that(result, equal_to(1))
+
+        result = self._tenant_dao.count(visible_tenants, domain_name='google.ca')
+        assert_that(result, equal_to(1))
+
+        result = self._tenant_dao.count(visible_tenants, domain_name='outlook.fr')
+        assert_that(result, equal_to(1))
+
+        result = self._tenant_dao.count(visible_tenants, search_domain='yahoo')
+        assert_that(result, equal_to(2))
+
+    @fixtures.db.tenant(name='baz a', domain_names=VALID_DOMAIN_NAMES_1)
+    @fixtures.db.tenant(name='bar b', domain_names=VALID_DOMAIN_NAMES_2)
+    @fixtures.db.tenant(name='foo c', domain_names=VALID_DOMAIN_NAMES_3)
     @fixtures.db.user()
     @fixtures.db.user()
     def test_list(self, a, b, c, user1_uuid, user2_uuid):
@@ -173,6 +187,14 @@ class TestTenantDAO(base.DAOTestCase):
 
         result = self._tenant_dao.list_(search='ba')
         expected = build_list_matcher('bar b', 'baz a')
+        assert_that(result, contains_inanyorder(*expected))
+
+        result = self._tenant_dao.list_(search_domain='yahoo')
+        expected = build_list_matcher('bar b', 'foo c')
+        assert_that(result, contains_inanyorder(*expected))
+
+        result = self._tenant_dao.list_(domain_name='outlook.fr')
+        expected = build_list_matcher('foo c')
         assert_that(result, contains_inanyorder(*expected))
 
         result = self._tenant_dao.list_(order='name', direction='desc')
@@ -252,10 +274,12 @@ class TestTenantDAO(base.DAOTestCase):
             raises(exceptions.UnknownTenantException),
         )
 
-    @fixtures.db.tenant()
+    @fixtures.db.tenant(domain_names=VALID_DOMAIN_NAMES_1)
     def test_delete_tenant_with_children(self, tenant_uuid):
 
-        a_uuid = self._create_tenant(name='a', parent_uuid=tenant_uuid)
+        a_uuid = self._create_tenant(
+            name='a', parent_uuid=tenant_uuid, domain_names=VALID_DOMAIN_NAMES_2
+        )
 
         assert_that(
             calling(self._tenant_dao.delete).with_args(tenant_uuid),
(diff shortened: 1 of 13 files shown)


nissou91:
recheck

nissou91:
recheck

wazo_auth/database/models.py
```diff
@@ -128,6 +130,23 @@ class Tenant(Base):
...
+class DomainName(Base):
+
+    __tablename__ = 'auth_tenant_domain_name'
+
+    name = Column(String(RFC_DN_MAX_LENGTH), nullable=False, unique=True)
+    tenant_uuid = Column(
+        String(38), ForeignKey('auth_tenant.uuid', ondelete='CASCADE'), nullable=False
+    )
+
+    uuid = Column(
```
pc-m, line 147: Could you move the `uuid` field before `name`. We usually put primary keys at the beginning of the schema definition.

wazo_auth/database/alembic/versions/0c9ccb1b16a8_add_auth_tenant_domain_name.py
```diff
@@ -0,0 +1,46 @@
...
+            nullable=False,
+        ),
+        sa.Column(
+            'name',
+            sa.String(RFC_DN_MAX_LENGTH),
+            nullable=False,
+            unique=True,
+        ),
+    )
+    op.create_unique_constraint('auth_tenant_domain_name_key', TABLE_NAME, ['name'])
```
pc-m, line 41: The `create_unique_constraint` is redundant with the `unique=True` in the `name` definition. You can remove it.

integration_tests/suite/test_db_tenant.py
```diff
@@ -299,24 +353,39 @@ def test_delete_access(self, tenant_uuid, policy_uuid):
         result = self.session.query(models.PolicyAccess).get(access_id)
         assert_that(result, equal_to(None))
 
-    def _assert_tenant_matches(self, uuid, name, parent_uuid=ANY_UUID, slug=ANY):
+    def _assert_tenant_matches(
+        self, uuid, name, parent_uuid=ANY_UUID, slug=ANY, domain_names=[]
```
pc-m, line 357: This is a python trap. You should not use an empty dict as a default argument. Use `None` instead.

integration_tests/suite/test_db_tenant.py
```diff
@@ -299,24 +353,39 @@ def test_delete_access(self, tenant_uuid, policy_uuid):
...
             .first()
         )
 
         assert_that(
-            tenant, has_properties(name=name, parent_uuid=parent_uuid, slug=slug)
+            tenant,
+            has_properties(name=name, parent_uuid=parent_uuid, slug=slug),
         )
 
+        if domain_names and len(domain_names) > 0:
```
pc-m, line 377: `if domain_names:` would be enough to not enter the if if `domain_names` was `None` or an empty list

integration_tests/suite/test_db_tenant.py
```diff
@@ -299,24 +353,39 @@ def test_delete_access(self, tenant_uuid, policy_uuid):
...
+
+            filter_ = models.DomainName.name.in_(domain_names)
+            count_ = s.query(models.DomainName).filter(filter_).count()
+            assert_that(count_, equal_to(len(domain_names)))
```
pc-m, line 381: This only check how many domain names there are, not if they match. You should do something similar to this
```suggestion
            names = s.query(models.DomainName.name).filter(filter_).all()
            assert_that(domain_names, contains_inanyorder(names))
```

wazo_auth/database/queries/tenant.py
```diff
@@ -75,12 +95,24 @@ def create(self, **kwargs):
         if uuid_:
             tenant.uuid = str(uuid_)
 
+        new_domain_names = []
+        if domain_names and len(domain_names) > 0:
```
pc-m, line 99: The marshmallow schema should set the `domain_names` to be an empty list and reject `None`. You can remove this `if`

wazo_auth/database/queries/tenant.py
```diff
@@ -50,9 +50,29 @@ def count_users(self, tenant_uuid, **kwargs):
 
         return self.session.query(User).filter(filter_).count()
 
+    def get_tenant_by_domain_name(self, domain_name):
```
pc-m, line 53: This function is unused, it should be removed with its test and the `_check_domain_name_exists` helper method

wazo_auth/plugins/http/tenants/tests/test_tenants.py
```diff
@@ -120,5 +126,109 @@ def test_that_empty_tenant_addresses_return_400(self, TenantDetector):
...
+                        '',
+                        [],
+                        {},
+                        ['wazo.io'],
+                    ]
+                },
+            ),
+        ]
+
+        for field, invalid_data in invalid_datas:
```
pc-m, line 161: This test does only 1 post with a body containing all possible wrong domain names in the same payload.
It should do a post for each of the possible value.

nissou91:
recheck

integration_tests/suite/test_tenants.py
```diff
@@ -443,6 +443,51 @@ def test_put(self, tenant, user):
...
+            has_entries(
+                uuid=tenant['uuid'],
+                name=name,
+                contact=user['uuid'],
+                address=has_entries(**ADDRESS_1),
+                domain_names=is_not(empty()),
+            ),
+        )
+
+        s = t.TenantDAO().session
```
pc-m, line 476: You can use a `GET` here and avoid the query

integration_tests/suite/test_tenants.py
```diff
@@ -224,7 +269,31 @@ def bus_received_msg():
...
+            '=wazo.io',
+            '+wazo.io',
+            '_wazo.io',
+            'wazo_io',
+            'wazo_io  ',
+            'x' * 62,
+        ]
+        for invalid_domain_name in invalid_domain_names:
+            assert_http_error(
+                400, self.client.tenants.new, domain_names=list(invalid_domain_name)
```
pc-m, line 293: I want to see some tests where `domain_names` is not a list.
`None`, `True`, `False`, `'wazo.community'`, `42`, `{'name': 'wazo.community'}`

wazo_auth/plugins/http/tenants/tests/test_tenants.py
```diff
@@ -120,5 +126,98 @@ def test_that_empty_tenant_addresses_return_400(self, TenantDetector):
                 invalid_data,
             )
 
+    @patch('wazo_auth.plugins.http.tenants.http.TenantDetector')
+    def test_that_tenant_with_invalid_domain_names_returns_400(self, TenantDetector):
+        TenantDetector.autodetect.return_value = Mock(uuid=s.tenant_uuid)
+
+        invalid_datas = [
```
pc-m, line 133: I want to see some tests where `domain_names` is not a list.
`None`, `True`, `False`, `'wazo.community'`, `42`, `{'name': 'wazo.community'}`

wazo_auth/schemas.py
```diff
@@ -136,9 +151,17 @@ class UserSessionListSchema(BaseListSchema):
...
-    sort_columns = ['name', 'slug']
+    sort_columns = ['name', 'slug', 'domain_name', 'domain_names', 'search_domain']
     default_sort_column = 'name'
-    searchable_columns = ['uuid', 'uuids', 'name', 'slug']
+    searchable_columns = [
+        'uuid',
+        'uuids',
+        'name',
+        'slug',
+        'domain_name',
```
pc-m, line 161: Only `domain_name` is required here.
The `searchable_columns` list contains fields that can be used in the GET (list) query string to search for a given tenant.
We want to add support for GET `/tenants?domain_name=wazo.community` and `/tenants?search=wazo`.
`search` is implicit.

wazo_auth/schemas.py
```diff
@@ -136,9 +151,17 @@ class UserSessionListSchema(BaseListSchema):
 
 
 class TenantListSchema(BaseListSchema):
-    sort_columns = ['name', 'slug']
+    sort_columns = ['name', 'slug', 'domain_name', 'domain_names', 'search_domain']
```
pc-m, line 154: The `sort_columns` is used to validate which columns can be used for the sort argument or a request. I don't think we should modify it. Sorting by domain_names would be weird since we can have multiple `domain_names` on a given tenant.

integration_tests/suite/test_tenants.py
```diff
@@ -267,35 +338,73 @@ def then(result, total=4, filtered=4, item_matcher=contains(top_tenant)):
             )
 
         result = self.client.tenants.list()
-        matcher = contains_inanyorder(foobaz, foobar, foobarbaz, top_tenant)
+        matcher = contains_inanyorder(
```
pc-m, line 341: Tests are telling you something here and you're telling the test to shutup about it.
The result of any request has the following format
```javascript
{
  'uuid': '94304416-2332-4617-bcae-c00f27e0411a',
  'parent_uuid': 'd5b20a36-0db2-484b-8587-d80a6a75af44',
  'name': 'foobaz',
  'slug': 'bbb',
  'contact': None,
  'phone': None,
  'domain_names': [
    '<wazo_auth.database.models.DomainName object at 0x7f5e90607850>', 
    '<wazo_auth.database.models.DomainName object at 0x7f5e90607b50>', 
    '<wazo_auth.database.models.DomainName object at 0x7f5e90607d50>'
  ],
  'address': {'line_1': None, 'line_2': None, 'city': None, 'state': None, 'country': None, 'zip_code': None}
}
```
The marshmallow schema should be modified to return the domain name instead of this. That will allow the test_post to work with a `contains_inanyorder` and avoid using `has_entries(uuid=` here to avoid comparing the other fields.

pc-m (changes requested):
Good job @nissou91
There some new fixes to make but it's mostly done.

wazo_auth/database/queries/tenant.py
```diff
@@ -231,6 +232,21 @@ def _generate_slug(self, name):
...
+    def _add_domain_filters(self, search_domain=None, domain_name=None, **ignored):
+        filter_ = text('true')
+        if search_domain:
+            search_pattern = '%{}%'.format(search_domain)
+            domain_name_search_filter_ = Tenant.domain_names.any(
+                DomainName.name.ilike(search_pattern)
+            )
+            filter_ = and_(filter_, domain_name_search_filter_)
+        if domain_name:
+            domain_name_strict_filter_ = Tenant.domain_names.any(
```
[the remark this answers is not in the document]
nissou91: Thanks, took me a while to find it :)

wazo_auth/schemas.py
nissou91: I have kept `search_domain` as looking for a specific keyword in domain names using `search` does not work; however, when I put `search_domain` (like: `list(search_domain='yahoo')`), `list_` and `count` work perfectly. Also, you have to add it to searchable_columns in order for `test_list` method in `test_tenants.py` to work correctly.

pc-m: > I have kept `search_domain` as looking for a specific keyword in domain names using `search` does not work; however, when I put `search_domain` (like: `list(search_domain='yahoo')`), `list_` and `count` work perfectly. Also, you have to add it to searchable_columns in order for `test_list` method in `test_tenants.py` to work correctly.

Yeah I missed it during the review, you should change `_add_domain_filters` to handle `search` and `domain_name` only.

Our HTTP API does not have ways to do a fuzzy search on a given field. The only option should be `domain_name` for a strict domain name match or search for a fuzzy search on the domain names and all other searchable fields

nissou91: could you please elaborate more your point? do you mean something like this?
```
def _add_domain_filters(self, search=None, domain_name=None, **ignored):
        if search:
            search_pattern = '%{}%'.format(search)
            domain_name_search_filter_ = Tenant.domains.any(
                DomainName.name.ilike(search_pattern)
            )
            return domain_name_search_filter_
        elif domain_name:
            domain_name_strict_filter_ = Tenant.domains.any(
                DomainName.name == domain_name
            )
            return domain_name_strict_filter_
        else:
            return text('true')
```

pc-m (approved):

PR #222, opened 2022-05-05
Why: fix regex for domain names

nissou91:


wazo_auth/database/alembic/versions/0c9ccb1b16a8_add_auth_tenant_domain_name.py
```diff
@@ -13,7 +13,7 @@
 down_revision = 'a6cda77a7e3f'
 
 TABLE_NAME = 'auth_tenant_domain'
-RFC_DN_MAX_LENGTH = 61
+RFC_DN_MAX_LENGTH = 253
```
pc-m, line 16: You cannot change the script because some people already upgraded there stack. Sylvain for example. You'll have to create a new alembic migration that changes the length of the name field.

pc-m (approved):

PR #225, opened 2022-05-31
Pull request for NES-1892-fix-slow-tenants

fblackburn1:
A best fix would be to only use ORM or only NOT use ORM. Mixing plain sql and ORM is not really friendly...


pc-m (approved):

fblackburn1:
recheck

fblackburn1:
recheck

fblackburn1:
recheck

sduthil:
recheck

sduthil:
recheck

fblackburn1:
regate

PR #227, opened 2022-06-08
config: accumulate lists from different config files

sduthil:
Why:

* This allows drop-in files to add access to default policies.


sduthil:
recheck

PR #228, opened 2022-06-14
Pull request for WAZO-2788-purpose-admins

sduthil:
## users: add purpose tenant_admin

Why:

* Can be used to distinguish users from admins inside a tenant

## users api: factorize purpose enum


sduthil:
We chose a different path for listing tenant admins: we shall list users based on the policy that they have access to (e.g. wazo_default_admin_policy)

PR #230, opened 2022-06-20
users: filter list by policy

sduthil:
Why:

* In order to get a list of users having a "role" (i.e. a policy)


sboily:
Please don't forget the CHANGELOG.md when you add something on the API.

sduthil:
recheck

fblackburn1 (approved):

sduthil:
regate

PR #242, opened 2022-08-02
use flask-restful parsing to validate input

fblackburn1:
why: improve robustness and avoid logging user input (for user_uuid)


pc-m (approved):

fblackburn1:
recheck

PR #245, opened 2022-08-17
LDAP authentification is case sensitive

nissou91:
**Problem:**
* When we have a user comfigured within ldap configuration with the following email `awonderland@wazo-auth.com`, and this user tries to login with the same email address but containing upper case letters instead; like: `AWONDERLAND@WAZO-AUTH.COM` then it raises an HTTP 401 error code.

**Expected Behaviour:**
* We should accept any email address as long as they are the same when lowered.


fblackburn1:
superset by https://github.com/wazo-platform/wazo-auth/pull/246

PR #246, opened 2022-08-17
LDAP authentication is case-sensitive

nissou91:
**Problem:**
* When we have a user comfigured within ldap configuration with the following email `awonderland@wazo-auth.com`, and this user tries to login with the same email address but containing upper case letters instead; like: `AWONDERLAND@WAZO-AUTH.COM` then it raises an HTTP 401 error code.

**Expected Behaviour:**
* We should accept any email address as long as they are the same when lowered; but we should not change/modify the current configuration or user creation process

### DIFF
--- a/wazo_auth/database/queries/user.py
+++ b/wazo_auth/database/queries/user.py
@@ -1,7 +1,7 @@
 # Copyright 2017-2022 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-from sqlalchemy import and_, or_, exc, func, text
+from sqlalchemy import and_, or_, exc, text
 from sqlalchemy.orm import joinedload
 from .base import BaseDAO, PaginatorMixin
 from . import filters
@@ -241,7 +241,7 @@ class UserDAO(filters.FilterMixin, PaginatorMixin, BaseDAO):
             .outerjoin(Email)
             .filter(
                 and_(
-                    func.lower(Email.address) == func.lower(login),
+                    Email.address == login,
                     Email.confirmed.is_(True),
                 )
             )
--- a/wazo_auth/database/queries/user.py
+++ b/wazo_auth/database/queries/user.py
@@ -1,7 +1,7 @@
 # Copyright 2017-2022 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-from sqlalchemy import and_, or_, exc, text
+from sqlalchemy import and_, or_, exc, func, text
 from sqlalchemy.orm import joinedload
 from .base import BaseDAO, PaginatorMixin
 from . import filters
@@ -239,7 +239,12 @@ class UserDAO(filters.FilterMixin, PaginatorMixin, BaseDAO):
         query = (
             self.session.query(User.uuid)
             .outerjoin(Email)
-            .filter(and_(Email.address == login, Email.confirmed.is_(True)))
+            .filter(
+                and_(
+                    func.lower(Email.address) == func.lower(login),
+                    Email.confirmed.is_(True),
+                )
+            )
         )
         row = query.first()
         if row:
--- a/wazo_auth/plugins/backends/ldap_user.py
+++ b/wazo_auth/plugins/backends/ldap_user.py
@@ -147,10 +147,10 @@ class LDAPUser(BaseAuthenticationBackend):
 
     def _get_user_by_ldap_attribute(self, user_email, tenant_uuid):
         for user in self._user_service.list_users(tenant_uuid=tenant_uuid):
-            if 'emails' in user.keys() and user['emails']:
-                wazo_auth_user_email = user['emails'][0]['address']
-                if wazo_auth_user_email.lower() == user_email.lower():
-                    return user
+            if user.get('emails'):
+                for email in user.get('emails'):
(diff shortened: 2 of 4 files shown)


nissou91:
recheck

wazo_auth/plugins/backends/tests/test_ldap_user.py
```diff
@@ -45,11 +45,12 @@ def setUp(self):
         self.tenant_service.get_by_uuid_or_slug.return_value = {
             'uuid': 'tenant-uuid-1234',
         }
+        self.wazo_email = 'wazo.io'
```
bloom1, line 48: I don't think it was necessary to extract this value in a variable - SonarCloud should be disabled (or consider it useless) for test files.

wazo_auth/plugins/backends/ldap_user.py
```diff
@@ -146,11 +146,11 @@ def _get_tenant(self, tenant_id):
         return self._tenant_service.get_by_uuid_or_slug(None, tenant_id)
 
     def _get_user_by_ldap_attribute(self, user_email, tenant_uuid):
-        for user in self._user_service.list_users(
-            email_address=user_email, tenant_uuid=tenant_uuid
-        ):
-            return user
-
+        for user in self._user_service.list_users(tenant_uuid=tenant_uuid):
```
bloom1, line 149: I would rather see this in the [list_](https://github.com/wazo-platform/wazo-auth/blob/WAZO-2852/wazo_auth/database/queries/user.py#L276) method. You can pretty much do like the other parameters extracted from the `kwargs` and do the lowercase comparison there. That would also allow the user search to be case insensitive when searching email addresses. In that case, you would also need to update the API documentation to reflect those changes.

bloom1 (changes requested):

wazo_auth/plugins/backends/tests/test_ldap_user.py
sboily: No no, you don't disable Sonarcloud :), you just need to update it to ignore this.

bloom1: > No no, you don't disable Sonarcloud :), you just need to update it to ignore this.

I think you misread what I wrote. I said that SonardCloud should be disabled for (to ignore) test files ;)

sboily: haha you right, the () has been considering a new phrase in my head. Sorry ;)

nissou91: How do you achieve that? :)

fblackburn1 (commented):
Why we change the whole wazo-auth implementation when the only requested change is for LDAP only?
Don't forget wazo-auth is used elsewhere ... and email address is unique with case sensitive.
Quick review of this PR add security issues

wazo_auth/plugins/backends/ldap_user.py
```diff
@@ -146,11 +146,11 @@ def _get_tenant(self, tenant_id):
...
+            if 'emails' in user.keys() and user['emails']:
```
bloom1, line 150: It is possible (and better) to use `if user.get('emails')` :smiley:

wazo_auth/database/queries/user.py
```diff
@@ -239,7 +239,12 @@ def get_user_uuid_by_login(self, login):
         query = (
             self.session.query(User.uuid)
             .outerjoin(Email)
-            .filter(and_(Email.address == login, Email.confirmed.is_(True)))
+            .filter(
+                and_(
+                    func.lower(Email.address) == func.lower(login),
+                    Email.confirmed.is_(True),
+                )
+            )
```
bloom1, line 247: We do not want to change the implementation for now - this will affect all of wazo-auth.
We should limit ourselves to the LDAP plugin for now, so we only want to lowercase what we get from the LDAP server.

wazo_auth/plugins/backends/ldap_user.py
```diff
@@ -146,11 +146,11 @@ def _get_tenant(self, tenant_id):
...
+                wazo_auth_user_email = user['emails'][0]['address']
+                if wazo_auth_user_email.lower() == user_email.lower():
```
bloom1, line 152: We should not lowercase the email address from wazo-auth. If we do, we might have a situation where a wazo-auth user has the email address user@domain and another user with User@domain. We would not know which one to match and maybe we would login the wrong user. Leaving the case sensitivity on the wazo-auth side will allow the differentiation.

bloom1 (changes requested):

wazo_auth/database/queries/user.py
nissou91: This will imply to remove this test scenario `test_ldap_authentication_works_when_case_sensitive_wazo_auth_user_email`; otherwise it will fail without this code change...

bloom1: > This will imply to remove this test scenario `test_ldap_authentication_works_when_case_sensitive_wazo_auth_user_email`; otherwise it will fail without this code change...

Exactly.
The wazo-auth user email is not case insensitive. The LDAP one is :smile:

wazo_auth/plugins/backends/ldap_user.py
nissou91: Done

PR #249, opened 2022-09-06
Inter-tenant association should fail

nissou91:


wazo_auth/exceptions.py
```diff
@@ -421,3 +421,16 @@ def __init__(self, name):
 
     def __str__(self):
         return f'Group "{self._name}" already exists'
+
+
+class UnauthorizedResourcesMutualAccessAttemptException(APIException):
+    def __init__(self, tenant_uuid, group_uuid):
```
pc-m, line 427: Could you rename the variables to `user_tenant_uuid` and `group_tenant_uuid`

pc-m (changes requested):
Looks good to me. Only one variable name change needed.

wazo_auth/exceptions.py
nissou91: You're right, it is done. Thanks.

pc-m (approved):

PR #254, opened 2022-10-19
WAZO-2899: `auth` publish on wazo-headers

ja0x:


fblackburn1 (changes requested):
This project must be taken separately
Any breaking change (i.e. `amqp` --> `bus`) must change in all projects
See with me for more information

ja0x:
Since modifications are more involved, this PR is closed in favor of a new one

PR #255, opened 2022-11-18
WAZO-2934: add `purpose` key into token's metadata

ja0x:


wazo_auth/plugins/metadata/default_user.py
```diff
@@ -17,5 +23,7 @@ def get_token_metadata(self, login, args):
             'auth_id': default_metadata['auth_id'],
             'pbx_user_uuid': default_metadata['uuid'],
             'xivo_uuid': default_metadata['xivo_uuid'],
+            'purpose': 'user',
+            'admin': self.is_admin(default_metadata['uuid']),
```
fblackburn1, line 27: Note that `admin` should be here **only** if the `wazo_default_admin_policy` is present on the DB
wazo-auth must stay generic, and not all implementation have this policy or admin concept

fblackburn1: It will be probably better to add a separate metadata plugin (but keep it inside wazo-auth for simplicity)

ja0x: Sure, definitely possible to do.  I would still keep `purpose` in the default, but move `admin` into a separate metadata plugin for the stack/pbx

fblackburn1 (approved):

PR #257, opened 2022-12-05
WAZO-2944 retrieve external provider connected users list

ja0x:
Depends-On: https://github.com/wazo-platform/wazo-auth-client/pull/32

### DIFF
--- a/integration_tests/suite/test_external.py
+++ b/integration_tests/suite/test_external.py
@@ -186,7 +186,9 @@ class TestExternalAuthAPI(base.ExternalAuthIntegrationTest):
             has_entries(
                 total=equal_to(1),
                 filtered=equal_to(1),
-                items=contains_inanyorder(session1['user_uuid']),
+                items=contains_inanyorder(
+                    has_entries(uuid=session1['user_uuid']),
+                ),
             ),
         )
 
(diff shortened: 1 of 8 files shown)


integration_tests/suite/test_external.py
```diff
@@ -173,6 +174,38 @@ def test_list(self, user1, user2, user3):
...
+            results = self.client.external.list_connected_users('foo', recurse=True)
+        finally:
+            client.token.revoke(token['token'])
+
+        assert_that(
+            results,
+            has_entries(
+                total=2,
+                filtered=2,
+                items=contains_inanyorder(session1['user_uuid'], user2['uuid']),
```
pc-m, line 205: You know what. In 3 months someone's going to want to use this resource to display the result in a UI. That person will want more information about the user but we will be locked in this result. To make it "future proof" I would change the result to
```javascript
{
  "total": 2,
  "filtered": 2,
  "items: [
    {"uuid": sesssion1['user_uuid']},
    {"uuid": user2['uuid']},
  ]
}
```

pc-m (changes requested):

integration_tests/suite/test_external.py
ja0x: I thought about it, but forgot to come back and change it! Thanks

ja0x:
recheck

pc-m (approved):

PR #265, opened 2023-01-17
[WP-749] ldap: add non-case-sensitive e-mail comparison wazo-auth side

bloom1:


pc-m (changes requested):
Could you add a test where we create multiple users

User 1: email = Foobar@example.com
User 2: email = FooBar@example.com
User 3: username = foobar@example.com

I would expect a 409 from the User 2 and User 3 creation as that would make the login unreliable for some of the users since only one would be check when creating a token.

bloom1:
> 

This will be fixed and tested in WP-750

wazo_auth/database/queries/user.py
```diff
@@ -232,16 +232,20 @@ def get_credentials(self, user_uuid):
             raise exceptions.UnknownUserUUIDException(user_uuid)
         return row.password_hash, row.password_salt
 
-    def get_user_uuid_by_login(self, login):
+    def get_user_uuid_by_login(self, login, case_sensitive=True):
```
sduthil, line 235: Do we really need a `case_sensitive` parameter?

sduthil (approved):

PR #269, opened 2023-02-02
openapi: fix inconsistency default values

fblackburn1:
also remove redundant example, now that default value is fixed ..


ja0x:
recheck

ja0x (approved):

PR #274, opened 2023-03-30
Add condition to not start the thread

franckrupin:


franckrupin:
recheck

pc-m (approved):

etc/wazo-auth/config.yml
```diff
@@ -5,6 +5,8 @@
 # Directory in which to place config overrides
 extra_config_files: /etc/wazo-auth/conf.d/
 
+update_policy_on_startup: True
```
fblackburn1, line 8: Is there a real use case for user of wazo-platform to have access to this option now?
IMHO we should not expose this option (i.e. do not put in the config.yml, which is the documentation for the user)
I would leave it internally only (hidden config), and it would avoid us to publicly maintain the upgrade of this option in the future

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 23.07
+
+* The following CLI options have been added
+  * `--listen-port` will override the HTTP server listening port
+  * `--log-file` will override the log filename
```
fblackburn1, line 7: Any reason to chose different name from the configuration file?
IMHO it add some confusion to remember that cli is `log_file` , but in configuration it's: `log_filename`
Same for `--listen-port` vs ` --rest-api-port` (but this one is debatable...)

PR #287, opened 2023-08-03
Authorization encoding

pc-m:


pc-m:
Closing because upgrading werkzeug fixes the issue

pc-m:
> Closing because upgrading werkzeug fixes the issue

The new version of werkzeug prevents a 500 from happening but does not work. We always get a 401

PR #289, opened 2023-08-09
Updated integration test dockerfile editable install command

DrPyser:
https://wazo-dev.atlassian.net/browse/WAZO-3294


jfbourque:
recheck

PR #297, opened 2023-12-12
token: validate the max value for expiration

pc-m:
super large values result in a psycopg2 error. based on sys.maxsize

I limited the expiration to 10 years which I guess is big enough for most people


franckrupin (approved):
LGTM

PR #302, opened 2024-01-15
WAZO-3600: prevent thread crashing when removing tokens without tenant_uuid

ja0x:
Why:

* Because the event requires a tenant_uuid, when none are provided (such as email reset token) the ExpiredTokenRemover thread crashes while trying to create it.


fblackburn1 (changes requested):
Missing test

fblackburn1 (changes requested):
Can you move test beside other expired token tests (`test_session.py`). It will avoid some confusion

ja0x:
Sure thing, didn't realize when I looked quickly there were some in test_sessions

fblackburn1 (approved):

PR #309, opened 2024-04-02
Pull request for WP-1321-tenant-performance

sduthil:
## performance tests: ensure db data is consistent with API

Why:

* This will raise an error if the API has a new side-effect in the
  database in the future
* We don't want to inject data that do not reflect reality

## tests: move performance tests to their own directory

Why:

* Avoid slowing down functional integration tests with possibly
  long-running performance tests

## improve performance for tenant creation

Why:

* The update_groups iterates over all existing tenants, which is
  unnecessary when creating a new tenant

## tox: separate integration tests from performance tests

### DIFF
--- a/integration_tests/performance_suite/conftest.py
+++ b/integration_tests/performance_suite/conftest.py
@@ -1,0 +1,1 @@
+../suite/conftest.py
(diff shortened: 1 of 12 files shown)


pc-m (approved):

sduthil:
Rebased following concurrent merge. Labeling with `auto-merge`

PR #313, opened 2024-04-08
Pull request for wazo-tox-performance-2

sduthil:
## tox: don't run performance tests with integration tests

Why:

* performance tests may be too heavy to run after every push

## zuul: add daily performance tests


sduthil:
In theory yes, in practice I won't have time to go further, closing. Thanks for the nudge

PR #314, opened 2024-04-11
Wp 1370 saml authentication

pc-m:


wazo_auth/plugins/http/saml/http.py
```diff
@@ -0,0 +1,118 @@
...
+        except VerificationError as err:
+            logger.warn("Verification error: %s", err)
+            resp = ServiceError(f"Verification error: {err}")
+            return resp(self.environ, self.start_response)
+        except SignatureError as err:
+            logger.warn("Signature error: %s", err)
+            resp = ServiceError(f"Signature error: {err}")
+            return resp(self.environ, self.start_response)
+        except Exception as err:
+            logger.error("SAML unexpected error: %s" % err)
```
pc-m, line 69: You should not do string formatting in a logger call. If the logger is configured not to log logger.error the string formatting will be done here. If you use `logger.error("SAML unexpected error: %s", err)` (`,` instead of `%`) the string formatting will be done by the logging module only when the line is going to be logged

wazo_auth/services/saml.py
```diff
@@ -0,0 +1,127 @@
...
+                )
+            except Exception as inst:
+                logger.error('Error during SAML client init')
+                logger.exception(inst)
+        else:
+            logger.warn(
+                'SAML config is missing, won\'t be able to provide SAML related services'
+            )
+
+    def _extractTuplesFromListOfStrings(self, ls):
```
pc-m, line 52: snake_case should be used instead of CamelCase

PR #321, opened 2024-05-15
Wp 1370 saml authentication rebased

DrPyser:


wazo_auth/services/saml.py
```diff
@@ -0,0 +1,146 @@
...
+    def get_user_login_and_remove_context(self, saml_session_id):
+        logger.debug('sessions %s', self._outstanding_requests)
+        for key in self._outstanding_requests:
+            if self._outstanding_requests[key].saml_session_id == saml_session_id:
+                reqid = key
+        session_data: Optional[SamlAuthContext] = self._outstanding_requests.get(reqid)
+        if session_data:
+            logger.debug('session_data : %s', session_data)
+            del self._outstanding_requests[reqid]
+            return session_data.login
```
DrPyser, line 135: The `saml_session_id` is assumed to exist and be found, but it may not(e.g. if I reload the `postacs.html`). If it is not, `reqid` will not be defined and there will be a `UnboundLocalError`(variable is not defined) at line 131.

```suggestion
        reqid = next((key for key, request in self._outstanding_requests.items() if request.saml_session_id == saml_session_id), None)
        if reqid is None:
            logger.debug('no outstanding request found for saml session %s', saml_session_id)
            return None
        # reqid was found, SamlAuthContext exists
        session_data = self._outstanding_requests[reqid]
        logger.debug('session_data : %s', session_data)
        del self._outstanding_requests[reqid]
        return session_data.login
```

wazo_auth/plugins/http/saml/api.yml
```diff
@@ -0,0 +1,112 @@
...
+      consumes:
+      - application/json
+      produces:
+      - application/json
+      description: |
+        Returns the information (SAML request and redirect location) that
+        the client needs to use to authenticate using the configured identity
+        provider. The redirection does not happen automatically, the client
+        has to use the `location` provided in the body of the response.
+        provider
```
pc-m, line 13: I think the `provider` here is a copy/paste error

pc-m: fixed

wazo_auth/services/saml.py
pc-m: fixed

DrPyser:
For the dependency on the `domain name` configuratio of a tenant.
- there is already a "domain name" configuration setting in portal which is actually http ingress configuration(https://portal.staging.wazo.cloud/#/pbx.master-http-ingress) and is actually requested to be a URL(e.g. `https://my-stack.my-domain.com`); this will be confusing for admins. Should this "tenant domain name" be an actual resolvable DNS name, or just a globally unique identifier for the tenant?
-since the SAML configuration is loaded by wazo-auth at startup, if the domain name is not yet defined on the tenant wazo-auth will need to be restarted for the config to be active; an event listener could listen to tenant update events and reload the configuration accordingly, otherwise this should be documented if it is viable to require a restart of wazo services for a configuration change

sboily:
> there is already a "domain name" configuration setting in portal which is actually http ingress configuration(https://portal.staging.wazo.cloud/#/pbx.master-http-ingress) and is actually requested to be a URL(e.g. https://my-stack.my-domain.com); this will be confusing for admins. Should this "tenant domain name" be an actual resolvable DNS name, or just a globally unique identifier for the tenant? Maybe the http ingress conf could be renamed("http ingress URL" would probably be more appropriate?)

It's for the API, it's very different, there is a domain nam configuration for each tenant. Please use this.

pc-m:
> > there is already a "domain name" configuration setting in portal which is actually http ingress configuration(https://portal.staging.wazo.cloud/#/pbx.master-http-ingress) and is actually requested to be a URL(e.g. https://my-stack.my-domain.com); this will be confusing for admins. Should this "tenant domain name" be an actual resolvable DNS name, or just a globally unique identifier for the tenant? Maybe the http ingress conf could be renamed("http ingress URL" would probably be more appropriate?)
> 
> It's for the API, it's very different, there is a domain nam configuration for each tenant. Please use this.

I used the tenant domain list for the validation on load to be sure that a migration is possible when we add the API to configure SAML.

@DrPyser we do not care about reloads at the moment, that's a problem we will tackle when we add the API to create the SAML configuration during runtime

PR #326, opened 2024-05-29
Wp 1427 saml resource error payload

jirkah:
The https://github.com/wazo-platform/wazo-auth-client/pull/49 needs to be reviewed/merged first.

Depends-On: https://github.com/wazo-platform/wazo-auth-client/pull/49

### DIFF
--- a/zuul.yaml
+++ b/zuul.yaml
@@ -38,15 +38,6 @@
           V5l5YH7RHL8S2x95JOrAeDarWLyTRiMkr/WasIru23RFjA154aQa1j+SN2KwaQ=
 
 - job:
-    name: auth-tox-integration
-    description: Run auth integration tests
-    parent: wazo-tox-integration-py39
-    pre-run: zuul.d/integration-pre-run.yaml
-    timeout: 10800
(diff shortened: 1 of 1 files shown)


integration_tests/suite/test_saml.py
```diff
@@ -0,0 +1,83 @@
...
+    def test_acs_missing_token_param(self) -> None:
+        assert_http_error_partial_body(
+            400,
+            self._expected_dict(
+                'RelayState and/or SAMLResponse', 'invalid-data', 'saml'
+            ),
+            self.client.saml.acs,
+            response='response',
+            token=None,
+        )
```
DrPyser, line 83: these cover the known error conditions. What about tests for successful conditions, or partially successful conditions?

DrPyser: I mean a test where both `saml_response` and `relay_state` are set(not None), for example.

DrPyser (dismissed):

DrPyser (approved):

pc-m (approved):

PR #339, opened 2024-06-11
Wazo 3789 saml global config

pc-m:


DrPyser:
closing because of branch name change

PR #347, opened 2024-06-19
WAZO-3769 saml global config

rbienvenu0:


rbienvenu0:
recheck

wazo_auth/services/saml.py
```diff
@@ -57,26 +57,34 @@ def __init__(self, config: Config, tenant_service: TenantService):
...
+            value = global_saml_config.get(key)
+            if not value:
+                missing_keys.append(key)
+
+        if missing_keys:
+            logger.error(
+                'Failed to initialize SAML service, missing configuration %s',
+                ','.join(missing_keys),
+            )
+            return
```
DrPyser, line 78: Is it desirable to leave the SAMLService partially initialized?
SAML authentication attemps will likely result in errors when trying to access `self._saml_clients[domain]`, do we want to have better error handling there?

wazo_auth/config.py
```diff
@@ -128,6 +128,7 @@
         'xmlsec_binary': '/usr/bin/xmlsec1',
         'key_file': '/var/lib/wazo-auth/saml/server.key',
         'cert_file': '/var/lib/wazo-auth/saml/server.crt',
+        'allow_unknown_attributes': True,
```
DrPyser, line 131: Does the behavior change with this new option?

pc-m: @sboily is this option necessary to get Google Workspace to work

DrPyser (dismissed):
Behavior seems generally good and consistent - saml config change is reflected in saml behavior(could not test `allow_unknown_attributes` yet)

wazo_auth/config.py
pc-m: I removed this option as it is not required at this time

DrPyser (approved):

PR #348, opened 2024-06-20
WP-1387 API to configure SAML

jirkah:
Depends-On: https://github.com/wazo-platform/wazo-auth-client/pull/52

### DIFF
--- a/wazo_auth/database/queries/saml_config.py
+++ b/wazo_auth/database/queries/saml_config.py
@@ -1,6 +1,8 @@
 # Copyright 2022-2024 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from typing import Any
+
 from sqlalchemy import exc
 
 from ... import exceptions
@@ -12,6 +14,15 @@ class SAMLConfigDAO(BaseDAO):
     def list(self) -> list[SAMLConfig]:
         return self.session.query(SAMLConfig).all()
 
+    def _forge_return(self, saml_config: SAMLConfig) -> dict[str, Any]:
+        return {
+            'domain_uuid': saml_config.domain_uuid,
+            'tenant_uuid': saml_config.tenant_uuid,
+            'entity_id': saml_config.entity_id,
+            'idp_metadata': saml_config.idp_metadata,
+            'acs_url': saml_config.acs_url,
+        }
+
     def get(self, tenant_uuid):
         saml_config = (
             self.session.query(SAMLConfig)
@@ -19,13 +30,7 @@ class SAMLConfigDAO(BaseDAO):
             .first()
         )
         if saml_config:
-            return {
-                'domain_uuid': saml_config.domain_uuid,
-                'tenant_uuid': saml_config.tenant_uuid,
-                'entity_id': saml_config.entity_id,
-                'idp_metadata': saml_config.idp_metadata,
-                'acs_url': saml_config.acs_url,
-            }
+            return self._forge_return(saml_config)
         raise exceptions.UnknownSAMLConfigException(tenant_uuid)
 
     def create(
@@ -35,7 +40,7 @@ class SAMLConfigDAO(BaseDAO):
         entity_id,
         idp_metadata,
         acs_url,
-    ):
+    ) -> dict[str, Any]:
         saml_config = SAMLConfig(
             tenant_uuid=tenant_uuid,
             domain_uuid=domain_uuid,
@@ -50,8 +55,17 @@ class SAMLConfigDAO(BaseDAO):
             self.session.rollback()
             if e.orig.pgcode == self._UNIQUE_CONSTRAINT_CODE:
                 raise exceptions.DuplicatedSAMLConfigException(tenant_uuid)
-            raise
-        return saml_config.tenant_uuid
+            if e.orig.pgcode == self._FKEY_CONSTRAINT_CODE:
+                raise exceptions.SAMLConfigParameterException(
+                    tenant_uuid,
+                    'Domain_uuid violates foreign key constraint, domain not in requested tenant',
+                    409,
+                )
+            else:
+                raise exceptions.SAMLConfigParameterException(
+                    tenant_uuid, f'Unexpected error on update {e.orig.pgcode}', 500
+                )
+        return self._forge_return(saml_config)
 
     def update(self, tenant_uuid, **kwargs):
         filter_ = SAMLConfig.tenant_uuid == str(tenant_uuid)
@@ -63,7 +77,9 @@ class SAMLConfigDAO(BaseDAO):
             self.session.flush()
         except exc.IntegrityError:
             self.session.rollback()
-            raise
+            raise exceptions.SAMLConfigParameterException(
+                tenant_uuid, 'Integrity error on update', '500'
+            )
 
     def delete(self, tenant_uuid) -> None:
         filter_ = SAMLConfig.tenant_uuid == str(tenant_uuid)
--- a/wazo_auth/database/queries/saml_config.py
+++ b/wazo_auth/database/queries/saml_config.py
@@ -5,7 +5,8 @@ from typing import Any
 
 from sqlalchemy import exc
 
-from ... import exceptions
+from wazo_auth import exceptions
+
 from ..models import SAMLConfig
 from .base import BaseDAO
 
--- a/wazo_auth/database/queries/saml_config.py
+++ b/wazo_auth/database/queries/saml_config.py
@@ -62,7 +62,7 @@ class SAMLConfigDAO(BaseDAO):
             self.session.rollback()
             raise
 
-    def delete(self, tenant_uuid):
+    def delete(self, tenant_uuid) -> None:
         filter_ = SAMLConfig.tenant_uuid == str(tenant_uuid)
         self.session.query(SAMLConfig).filter(filter_).delete(synchronize_session=False)
         self.session.flush()
--- a/wazo_auth/database/queries/saml_config.py
+++ b/wazo_auth/database/queries/saml_config.py
@@ -9,6 +9,9 @@ from .base import BaseDAO
 
 
 class SAMLConfigDAO(BaseDAO):
+    def list(self) -> list[SAMLConfig]:
+        return self.session.query(SAMLConfig).all()
+
     def get(self, tenant_uuid):
         saml_config = (
             self.session.query(SAMLConfig)
@@ -17,6 +20,7 @@ class SAMLConfigDAO(BaseDAO):
         )
         if saml_config:
             return {
+                'domain_uuid': saml_config.domain_uuid,
                 'tenant_uuid': saml_config.tenant_uuid,
                 'entity_id': saml_config.entity_id,
                 'idp_metadata': saml_config.idp_metadata,
@@ -26,11 +30,13 @@ class SAMLConfigDAO(BaseDAO):
     def create(
         self,
         tenant_uuid,
+        domain_uuid,
         entity_id,
         idp_metadata,
     ):
         saml_config = SAMLConfig(
             tenant_uuid=tenant_uuid,
+            domain_uuid=domain_uuid,
             entity_id=entity_id,
             idp_metadata=idp_metadata,
         )
--- a/wazo_auth/database/queries/saml_config.py
+++ b/wazo_auth/database/queries/saml_config.py
@@ -1,0 +1,66 @@
+# Copyright 2022-2024 The Wazo Authors  (see the AUTHORS file)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from sqlalchemy import exc
+
+from ... import exceptions
+from ..models import SAMLConfig
+from .base import BaseDAO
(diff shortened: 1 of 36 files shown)


wazo_auth/database/queries/saml_config.py
```diff
@@ -0,0 +1,72 @@
...
+            raise
+
+    def delete(self, tenant_uuid) -> None:
+        filter_ = SAMLConfig.tenant_uuid == str(tenant_uuid)
+        self.session.query(SAMLConfig).filter(filter_).delete(synchronize_session=False)
+        self.session.flush()
+
+    def exists(self, tenant_uuid):
+        filter_ = SAMLConfig.tenant_uuid == str(tenant_uuid)
+        return self.session.query(SAMLConfig).filter(filter_).count() > 0
```
DrPyser, line 72: sql `EXISTS` with `.exists()` should produce optimal code for this use case.

wazo_auth/database/queries/saml_config.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2022-2024 The Wazo Authors  (see the AUTHORS file)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from sqlalchemy import exc
+
+from ... import exceptions
```
DrPyser, line 6: my opinion: relative imports hurt readability(complexify the job of understanding module dependencies).
Beyond the second degree of separation(`..`), I think we should stick to absolute paths, here `wazo_auth.exceptions`.

wazo_auth/plugins/http/saml_config/http.py
```diff
@@ -0,0 +1,130 @@
...
+    def get(self):
+        scoping_tenant: Tenant = Tenant.autodetect()
+        try:
+            if url := self._saml_config_service.get_acs_url(scoping_tenant.uuid):
+                return saml_acs_url_schema.dump(url), 200
+        except exceptions.APIException as e:
+            raise e
+        except Exception as e:
+            logger.exception(
+                f"An error occurred while getting SAML ACS URL for tenant {scoping_tenant.uuid}."
```
DrPyser, line 126:
```suggestion
                f"An error occurred while getting SAML ACS URL for tenant %s.", scoping_tenant.uuid
```

as a quirk of the logging system in python stdlib, it is preferred to write format strings using `%s`/ `%d` placeholders, and provide the format values as additional positional arguments after the format string. This allows the logging system to process the format string and the arguments, in principle allowing custom logic to make use of those separate pieces.

 I think this reflects existing conventions at wazo.

wazo_auth/plugins/http/saml_config/schemas.py
```diff
@@ -0,0 +1,23 @@
...
+    domain_uuid = fields.String(validate=Length(min=36, max=38), required=True)
+    entity_id = fields.String(validate=Length(min=1, max=512), required=True)
+    tenant_uuid = fields.String(validate=Length(min=36, max=38), required=False)
+
+
+saml_config_schema = SamlConfig()
+
+
+class SamlAcsUrl(BaseSchema):
+    acsrl = fields.String(validate=Length(min=1, max=2083), required=True)
```
DrPyser, line 20: typo?
```suggestion
    acs_url = fields.String(validate=Length(min=1, max=2083), required=True)
```

wazo_auth/services/saml.py
```diff
@@ -51,18 +51,46 @@ class Config(TypedDict, total=False):
...
+        tenant_service: TenantService,
+    ):
         self._config = config
         self._outstanding_requests: dict[Any, SamlAuthContext] = {}
         self._saml_clients: dict[str, Saml2Client] = {}
-        self._tenant_service = tenant_service
+        self._tenant_service: TenantService = tenant_service
 
-        self._init_clients()
+        # self._init_clients()
```
DrPyser, line 64:
```suggestion
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+        return etree_metadata
+
+    def get_acs_url(self, tenant_uuid: str) -> dict[str, str]:
+        return {'acsUrl': 'http://localhost:9497/api/auth/v1/saml/acs'}
+
+    def _update_domain_name(self, item, domains) -> dict[str, str]:
+        domain_name: list[str] = [
+            domain.name for domain in domains if domain.uuid == item['domain_uuid']
+        ]
+        if domain_name[0]:
```
DrPyser, line 71: indexing empty array will result in `IndexError` exception.
```suggestion
        if domain_name:
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+            item['domain_name'] = domain_name[0]
+            return item
+        assert f'Database consistency error, domain name for {item}/{domains}'
```
DrPyser, line 74: if the intent is to always yield an AssertionError at this point:
```suggestion
        assert False, f'Database consistency error, domain name for {item}/{domains}'
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+        return {}
+
```
DrPyser, line 76: if the intent is to always produce an AssertionError at this point
```suggestion
        assert False, f'Database consistency error, domain name for {item}/{domains}'

```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+    def _update_item(self, item, domains) -> dict[str, str]:
+        item = self._update_domain_name(item, domains)
+        item = self._add_metadata_path(item)
+        return item
+
+    def _reload_saml_service(self) -> None:
+        db_configs: list[SAMLConfig] = self._dao.saml_config.list()
+        domains: list[Domain] = self._dao.domain.list()
+        saml_configs = [saml_config_schema.dump(item) for item in db_configs]
```
DrPyser, line 89: what is the goal/value of using marshmallow schema here?

I find it weird to use marshmallow schema to process data in internal logic. 
They are designed to process data at system boundaries, deserializing and validating data from the outside(API inputs) and serializing data to expose to the outside(API outputs).
For data from database, we already know its structure and should be able to assume it to be valid.

Also, marshmallow schemas don't expose typed interface, so we don't know what data we're getting and passing on, as far as type goes its just a dict or a list of dicts.

In this case, why not keep using the `SAMLConfig` sqlalchemy objects provided by the DAO, which at least have interfaces with known attributes?

DrPyser, line 89:
```suggestion
        saml_configs = [cast(SAMLDomainConfigDict, saml_config_schema.dump(item)) for item in db_configs]
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+            item['domain_name'] = domain_name[0]
+            return item
+        assert f'Database consistency error, domain name for {item}/{domains}'
+        return {}
+
+    def _add_metadata_path(self, item) -> dict[str, str]:
+        item['metadata_path'] = self._get_metadata_path(item['tenant_uuid'])
+        return item
+
+    def _update_item(self, item, domains) -> dict[str, str]:
```
DrPyser, line 81: Assuming
```
class SAMLDomainConfigDict(TypedDict):
    domain_uuid: str
    entity_id: str
    tenant_uuid: str
```

```suggestion
    def _update_item(self, item: SAMLDomainConfigDict, domains: list[Domain]) -> dict[str, str]:
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+        domain_name: list[str] = [
+            domain.name for domain in domains if domain.uuid == item['domain_uuid']
+        ]
+        if domain_name[0]:
+            item['domain_name'] = domain_name[0]
+            return item
+        assert f'Database consistency error, domain name for {item}/{domains}'
+        return {}
+
+    def _add_metadata_path(self, item) -> dict[str, str]:
```
DrPyser, line 77:
```suggestion
    def _add_metadata_path(self, item: SAMLDomainConfigDict) -> dict[str, str]:
```

wazo_auth/services/saml_config.py
DrPyser: I would also change the method name to something more descriptive.

i think I understand this transforms db data to a domain config, so how about `_make_domain_config_from_db`

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+    def get_metadata(self, tenant_uuid: str) -> ElementTree.Element:
+        etree_metadata: ElementTree.Element = ElementTree.fromstring(
+            self._dao.saml_config.get(tenant_uuid)['idp_metadata']
+        )
+        return etree_metadata
+
+    def get_acs_url(self, tenant_uuid: str) -> dict[str, str]:
+        return {'acsUrl': 'http://localhost:9497/api/auth/v1/saml/acs'}
+
+    def _update_domain_name(self, item, domains) -> dict[str, str]:
```
DrPyser, line 67: See other comments for `SAMLDBDomainConfigEntryDict`

```suggestion
    def _update_domain_name(self, item: SAMLDBDomainConfigEntryDict, domains: list[Domain]) -> dict[str, str]:
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, config, saml_service, dao) -> None:
+        self._xml_files_dir: str = config['saml']['xml_files_dir']
+        self._saml_service = saml_service
+        super().__init__(dao)
+        self._reload_saml_service()
+
+    def _get_metadata_path(self, tenant_uuid: str) -> str:
+        return self._xml_files_dir + '/' + tenant_uuid + '.xml'
+
+    def _update_xml_metadata(
```
DrPyser, line 25:
```suggestion
    def _write_xml_metadata_to_filesystem(
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+        self, tenant_uuid: str, etree_metadata: ElementTree.ElementTree
+    ) -> None:
+        etree_metadata.write(self._get_metadata_path(tenant_uuid))
+
+    def _delete_conf(self, tenant_uuid: str) -> None:
+        os.remove(self._xml_files_dir + '/' + tenant_uuid + '.xml')
```
DrPyser, line 31: Reuse `_get_metadata_path` for the path(avoid duplication of logic, guarantee correctness)
```suggestion
        os.remove(self._get_metadata_path)
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SAMLConfigService(BaseService):
+    def __init__(self, config, saml_service, dao) -> None:
+        self._xml_files_dir: str = config['saml']['xml_files_dir']
+        self._saml_service = saml_service
+        super().__init__(dao)
+        self._reload_saml_service()
```
DrPyser, line 20: I see no reason not to call super first
```suggestion
        super().__init__(dao)
        self._xml_files_dir: str = config['saml']['xml_files_dir']
        self._saml_service = saml_service
        self._reload_saml_service()
```

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+from wazo_auth.database.models import Domain, SAMLConfig
+from wazo_auth.plugins.http.saml_config.schemas import saml_config_schema
+from wazo_auth.services.helpers import BaseService
+
+logger = logging.getLogger(__name__)
+
+
+class SAMLConfigService(BaseService):
+    def __init__(self, config, saml_service, dao) -> None:
+        self._xml_files_dir: str = config['saml']['xml_files_dir']
```
DrPyser, line 17: `xml_files_dir` is a new config attribute that is not otherwise declared in `wazo_auth.services.saml.WazoSAMLConfig`. Should be added there.

wazo_auth/services/saml_config.py
```diff
@@ -0,0 +1,94 @@
...
+
+from wazo_auth.database.models import Domain, SAMLConfig
+from wazo_auth.plugins.http.saml_config.schemas import saml_config_schema
+from wazo_auth.services.helpers import BaseService
+
+logger = logging.getLogger(__name__)
+
+
+class SAMLConfigService(BaseService):
+    def __init__(self, config, saml_service, dao) -> None:
```
DrPyser, line 16: Assuming additional imports
```
from wazo_auth.database.queries import DAO
from wazo_auth.services.saml import Config, SAMLService
```

```suggestion
    def __init__(self, config: Config, saml_service: SAMLService, dao: DAO) -> None:
```

pc-m (changes requested):
I spotted two missing tests. Charles also added some comments. I'd also like to be sure the the saml functional test keeps working since its not being executed by the CI at the moment

pc-m (approved):

chartrandf:
👀 (watch mode)

pc-m (approved):

PR #351, opened 2024-07-04
Wp 1526 wazo auth cli native authn

pc-m:


wazo_auth/database/alembic/versions/1cd6161394d0_native_user.py
```diff
@@ -0,0 +1,37 @@
...
+def upgrade():
+    query = (
+        user_table.update()
+        .values(authentication_method='native')
+        .where(
+            sa.or_(
+                user_table.c.purpose == 'internal',
+                user_table.c.purpose == 'external_api',
+            )
+        )
```
DrPyser, line 31: What if an administrator created an external_api user that uses saml auth?

pc-m: The administrator would have to change the authentication method for that user from `native` to `saml` or `default` if the tenant uses `saml`.

pc-m: https://github.com/wazo-platform/wazo-platform.org/pull/726 I wrote an upgrade note

DrPyser (approved):

PR #360, opened 2024-08-01
WP-1533 sp initiated logout

jirkah:
Depends-On: https://github.com/wazo-platform/wazo-auth-client/pull/53

### DIFF
--- a/wazo_auth/config.py
+++ b/wazo_auth/config.py
@@ -127,7 +127,7 @@ _DEFAULT_CONFIG = {
     'saml': {
         'domains': {},
         'acs_url_template': 'https://{{STACK_URL}}/api/auth/0.1/saml/acs',
-        'saml_session_lifetime_seconds': 600,
+        'saml_session_lifetime_seconds': 86400,
         'xmlsec_binary': '/usr/bin/xmlsec1',
         'key_file': '/var/lib/wazo-auth/saml/server.key',
         'cert_file': '/var/lib/wazo-auth/saml/server.crt',
--- a/wazo_auth/config.py
+++ b/wazo_auth/config.py
@@ -127,6 +127,7 @@ _DEFAULT_CONFIG = {
     'saml': {
         'domains': {},
         'acs_url_template': 'https://{{STACK_URL}}/api/auth/0.1/saml/acs',
+        'saml_login_timeout_seconds': 600,
         'saml_session_lifetime_seconds': 86400,
         'xmlsec_binary': '/usr/bin/xmlsec1',
         'key_file': '/var/lib/wazo-auth/saml/server.key',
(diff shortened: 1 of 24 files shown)


wazo_auth/database/models.py
```diff
@@ -212,6 +212,11 @@ class Token(Base):
     user_agent = Column(Text)
     remote_addr = Column(Text)
     acl = Column(ARRAY(Text), nullable=False, server_default='{}')
+    refresh_token_uuid = Column(
```
pc-m, line 215: Should be `String(36)`

wazo_auth/database/models.py
```diff
@@ -463,3 +468,7 @@ class SAMLSession(Base):
     relay_state = Column(String(44), nullable=False)
     login = Column(String(512), nullable=True)
     start_time = Column(DateTime(timezone=True), nullable=True)
+    saml_name_id = Column(Text, nullable=True)
+    refresh_token_uuid = Column(
```
pc-m, line 472: the field is named `refresh_token_uuid` and it refers to the `auth_tenant.uuid` column

wazo_auth/database/alembic/versions/2903ea6f2f20_add_refresh_token_uuid_to_auth_token.py
```diff
@@ -0,0 +1,32 @@
...
+table = 'auth_token'
+column_name = 'refresh_token_uuid'
+
+
+def upgrade():
+    op.add_column(
+        table,
+        sa.Column(
+            column_name,
+            sa.String(38),
```
pc-m, line 24: Should be `String(36)`

wazo_auth/database/alembic/versions/26146ddb2111_add_name_id_and_token_uuid_to_saml_.py
```diff
@@ -0,0 +1,39 @@
...
+            'saml_name_id',
+            sa.Text,
+            nullable=True,
+        ),
+    )
+    op.add_column(
+        table,
+        sa.Column(
+            'refresh_token_uuid',
+            sa.Text,
```
pc-m, line 31: The type should be the same as the model. You also need to add the FK constraint

pc-m (changes requested):

pc-m:
recheck

pc-m:
recheck

wazo_auth/config.py
```diff
@@ -126,7 +126,7 @@
     'saml': {
         'domains': {},
         'acs_url_template': 'https://{{STACK_URL}}/api/auth/0.1/saml/acs',
-        'saml_session_lifetime_seconds': 600,
+        'saml_session_lifetime_seconds': 86400,
```
pc-m, line 129: Should we introduce another value here?
We want to be able to do the logout a long time after the login but I wonder if it's safe to keep a session that has not been "consumer" for that long

pc-m (approved):

pc-m:
regate

PR #361, opened 2024-08-02
WAZO-2542: process expired tokens/sessions in batches

ja0x:
This PR modifies the ExpiredTokenRemover to process expired tokens and sessions in batches.  

The goal is prevent OOM errors when deleting too many tokens at once.  It will now process tokens and sessions by batches (5000 by default, configurable in config) until all expired tokens/sessions are removed.

### DIFF
--- a/wazo_auth/database/queries/token.py
+++ b/wazo_auth/database/queries/token.py
@@ -4,6 +4,10 @@
 import json
 import time
 
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import Bundle
+from sqlalchemy.sql import cast
+
 from ... import exceptions
 from ..models import Session, Tenant
 from ..models import Token as TokenModel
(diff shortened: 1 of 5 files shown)


pc-m (approved):

wazo_auth/database/queries/token.py
```diff
@@ -78,61 +82,64 @@ def delete(self, token_uuid):
...
+            uuids = [item['uuid'] for item in items]
+            self.session.query(model).filter(model.uuid.in_(uuids)).delete(
+                synchronize_session=False
+            )
+            self.session.flush()
+
+        while batch := self._get_tokens_and_sessions_by_expiration(0, batch_size):
+            tokens, sessions = batch
+            delete_by_uuids(TokenModel, tokens)
+            delete_by_uuids(Session, sessions)
```
fblackburn1, line 129: @ja0x Out of curiosity
1 .what was the performance impact of leaving the foreign key `session_uuid` delete the token?
  * If really necessary, maybe a comment to explain why this foreignkey bypass

2. What was the complexity to add a test in `performance_suite` about token deletion?

ja0x: 1. You mean using the cascade delete from session -> tokens?  No idea, blind oversight from copying/adapting previous code.  Can/should be implemented in another small PR

2. No complexity, felt like it didn't belong in `performance_suite` since the batching is configurable, there is no need to create thousands of tokens to verify it works.  The functionality of the batching is [tested here](https://github.com/wazo-platform/wazo-auth/pull/361/files/9dca4e7cefd734fc95a6124cc23486a72f38ab4d#diff-9e917df3a175e30420cb0a68b64da831dee2778e50c716c22b8ae10dca040df5R117-R157) and the previous tests already covers the token/session deletion

fblackburn1: Thanks I created https://wazo-dev.atlassian.net/browse/WAZO-3885 to avoid to forget it

PR #362, opened 2024-08-06
WP-1538 refuse password update/reset for users with saml/ldap auth

jirkah:


wazo_auth/plugins/http/password_reset/api.yml
```diff
@@ -25,6 +25,8 @@ paths:
       responses:
         '204':
           description: A mail will be sent if the username or email were found
+        '405':
```
pc-m, line 28: I would not return a `405` in the `GET`
A malicious user could use this URL to discover valid usernames and email addresses. Returning the `405` in the `POST` should be enough

pc-m (changes requested):

pc-m (approved):

pc-m:
regate

PR #371, opened 2024-09-13
test: use latest wazo-base-db image to pass params

fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-manage-db/pull/273
(it won't work until the Depends-on PR will be merged)

why: postgres command is now an entrypoint and parameter can be passed
as command


pc-m (approved):

fblackburn1:
recheck

fblackburn1:
recheck (unrelated fail)

PR #372, opened 2024-09-17
Wazo 3886 saml lougout fails after restart

jirkah:


wazo_auth/database/queries/saml_pysaml2_cache.py
```diff
@@ -0,0 +1,161 @@
...
+    ):
+        if not entities:
+            try:
+                cni: str = code(name_id)
+                entities = (
+                    self.session.query(SAMLPysaml2Cache)
+                    .filter(SAMLPysaml2Cache.name_id == cni)
+                    .all()
+                )
+            except Exception:
```
DrPyser, line 53: We should log an error(with stacktrace) here.
If we expect specific exceptions that should not be treated as errors(i.e. could occur normally), we should have specific except clauses for those.

wazo_auth/database/queries/saml_pysaml2_cache.py
```diff
@@ -0,0 +1,161 @@
...
+                try:
+                    tmp = set(res[key]).union(set(vals))
+                    res[key] = list(tmp)
+                except KeyError:
+                    res[key] = vals
+        return res, oldees
+
+    def get(self, name_id: NameID, entity_id: str, check_not_on_or_after: bool = True):
+        cni = code(name_id)
+        o = self._search(name_id=cni, entity_id=entity_id)
```
DrPyser, line 82: alternative to single letter variable? `cache_entry`?

wazo_auth/database/queries/saml_pysaml2_cache.py
```diff
@@ -0,0 +1,161 @@
...
+        if not o:
+            return None
+        timestamp = o[0].not_on_or_after
+        info = o[0].info
+        if not timestamp:
+            return None
+
+        info = json.loads(info).copy()
```
DrPyser, line 90: why do we copy right after instantiating a new object?

wazo_auth/database/queries/saml_pysaml2_cache.py
```diff
@@ -0,0 +1,161 @@
...
+        return [k.entity_id for k in self._search(name_id=cni)]
+
+    def receivers(self, name_id: NameID):
+        return self.entities(name_id)
+
+    def active(self, name_id: NameID, entity_id: str):
+        try:
+            cni = code(name_id)
+            entry = self._search(name_id=cni, entity_id=entity_id)[0]
+        except KeyError:
```
DrPyser, line 152: where do we expect the `KeyError` to come from?
If it's only one of the two line, I would only put the `try` around that line.

I expected the `[0]` indexing to potentially raise a `IndexError` instead.

wazo_auth/services/saml.py
```diff
@@ -175,9 +175,15 @@ def init_clients(self, db_configs):
                     'SAML config for domain: %s: %s', domain_name, raw_saml_config
                 )
                 try:
+                    from wazo_auth.database.queries.saml_pysaml2_cache import (
+                        SAMLPysaml2CacheDAO,
+                    )
```
DrPyser, line 180: why do we need to dynamically import this?

wazo_auth/services/saml.py
```diff
@@ -347,7 +354,10 @@ def get_user_login(self, saml_session_id: str) -> str | None:
         for reqid, session_data in self._dao.saml_session.list(
             session_id=saml_session_id
         ):
-            return session_data.login if session_data else None
+            try:
+                return session_data.login
+            except AttributeError:
+                return None
```
DrPyser, line 360: does this not deserve an error log? Is this a normal situation to not find the `login` attribute as expected?

DrPyser (approved):
fix validated, well done!

PR #376, opened 2024-10-14
WP-1654 protect against incomplete saml logout

jirkah:


wazo_auth/config.py
```diff
@@ -127,7 +127,7 @@
         'domains': {},
         'acs_url_template': 'https://{{STACK_URL}}/api/auth/0.1/saml/acs',
         'saml_login_timeout_seconds': 600,
-        'saml_session_lifetime_seconds': 86400,
+        'saml_session_lifetime_seconds': 604800,
```
DrPyser, line 130: why do we go from 60 days to 420 days? 60 days seems more reasonable(the smaller the default, the most secure)?

wazo_auth/services/saml.py
```diff
@@ -381,23 +378,34 @@ def update_refresh_token(self, refresh_token: UUID, saml_session_id: str) -> Non
...
+            elif now > expire_at and context.saml_session_id != 'token_already_used':
+                logger.debug("Deleting SAML session on timeout: %s", item)
+                self._dao.saml_session.delete(item.request_id)
+
+    def _clean_pysaml2_sessions(self, now: datetime) -> None:
+        one_week_ago: datetime = (
+            datetime.now(tz=timezone.utc) - self._saml_session_lifetime
+        )
+        one_week_ago_timestamp: int = int(round(one_week_ago.timestamp()))
+        for item in self._dao.saml_pysaml2_cache.get_expired(one_week_ago_timestamp):
```
DrPyser, line 401: since the session lifetime is variable and not constant, why is it assumed that this yields a date "one week ago"?

wazo_auth/config.py
DrPyser: Ah oups I messed up my calculation :facepalm:

tox.ini
```diff
@@ -30,7 +30,7 @@ pass_env =
     WAZO_TEST_NO_DOCKER_COMPOSE_PULL
     WAZO_TEST_PROFILING_DIR
 commands =
-    make test-setup
+    #make test-setup
```
DrPyser, line 33:
```suggestion
    make test-setup
```

DrPyser (approved):
LGTM/tested, thanks @jirkah !

DrPyser (approved):
LGTM, thanks @jirkah !

PR #386, opened 2025-03-05
fetch all users groups only once

fblackburn1:
why: improve startup speed
On an instance with 30k tenants, the old implementation took 10 seconds.
With the new implementation, it's instantaneous.

### DIFF
--- a/wazo_auth/services/all_users.py
+++ b/wazo_auth/services/all_users.py
@@ -1,4 +1,4 @@
-# Copyright 2020-2024 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2020-2025 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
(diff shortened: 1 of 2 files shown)


wazo_auth/services/all_users.py
```diff
@@ -30,12 +30,19 @@ def update_policies(self, tenant_uuids):
...
+            group.tenant_uuid: group for group in all_users_groups
+        }
+
         for tenant_uuid in tenant_uuids:
             self.associate_policies_for_tenant(
                 tenant_uuid,
                 policies,
                 current_group_policy_associations,
                 existing_config_managed_policies_by_tenant,
+                all_users_groups_by_tenant,
```
rbienvenu0, line 45: I believe it would be better to only pass the list of users groups for the current tenant instead of the entire dict. That way `associate_policies_for_tenant()` only have to have the argument `all_users_group`
```suggestion
                all_users_groups_by_tenant[tenant_uuid],
```

fblackburn1: Agreed, thank you :+1:

PR #394, opened 2025-05-26
WP 1841 refresh token interoperability

DrPyser:
- **integration_tests: test refresh token for custom idp**
- **controller: reorder idp loading**
- **idp: add get_backend method to idp interface**
- **token: debug log authentication error**


bloom1 (approved):
LGTM to me. That's simpler than I thought :smiley:

PR #403, opened 2025-08-13
WAZO 4218 idempotent update external mobile

DrPyser:
Depends-On: https://github.com/wazo-platform/wazo-bus/pull/126


DrPyser:
- fixed error in api spec (no nullable in openapi v2, but redocly supports `x-nullable`)
- changed tenants integrity error check to use psycopg2 errorcodes module constants instead

PR #415, opened 2025-10-03
fix-performance-log

fblackburn1:
- **test: remove unused condition**
- **test: scope asset by package to fix log override**


DrPyser (dismissed):
LGTM!

PR #424, opened 2025-12-23
WAZO-4352: Force json decoding requests content

rbienvenu0:
Why: forces to decode request content as json even if the content type isn't defined

WAZO-4352

---

> [!NOTE]
> Standardizes JSON body handling by forcing JSON parsing even when Content-Type is missing and validates that empty bodies are rejected.
> 
> - Replace `request.get_json()` with `request.get_json(force=True)` across endpoints: external auth (`google`, `microsoft`, `mobile`), config patch, external config (`post`/`put`), groups (`post`/`put`), IDP users (`put`), password reset (`post`), SAML SSO (`post`), tenants (`post`/`put`), user email update (`put`), user registration (`post`), users (`post`/`put`/password `put`)
> - Integration tests: add `_make_http_request` helper and `assert_empty_body_returns_400`; new tests verify `400` on empty bodies for groups, tenants, users (create/update/external), tokens, policies, config, SAML SSO, IDP SAML users, LDAP backend, Google external config, and password reset
> - Minor: update copyright years
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0bb8218c56fd1d3c1387510f470c4ad694bcf94a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/integration_tests/suite/test_http_interface.py
+++ b/integration_tests/suite/test_http_interface.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2025 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2015-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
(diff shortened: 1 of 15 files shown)


DrPyser (changes requested):

integration_tests/suite/test_http_interface.py
```diff
@@ -144,17 +144,116 @@ def test_that_no_type_returns_400(self):
...
+            except requests.HTTPError:
+                pass
+            try:
+                self.client.policies.delete(test_policy['uuid'])
+            except requests.HTTPError:
+                pass
+            try:
+                self.client.token.revoke(test_token)
+            except requests.HTTPError:
+                pass
```
DrPyser: - we want to avoid testing many independent things in the same test. Each of those endpoint have their own independent implementation, and can fail to meet the expectation independently. We want to see all those errors separately; testing in bulk like this would allow one failure to hide others;
- testing POST (create) and PUT (update) require different setup logic, so should be tested separately; just like different endpoints, different methods have separate implementation logic, and failures should be identified separately
- doing all this together makes this test heavy and hard to read; ideally tests should be small (in line count and in scope) and simple, and isolated; 
  sometimes we can test a class of expectations together when the testing logic lends to it (see confd), but this should not come at the cost of readability, simplicity and testing granularity;
- extracting the generic logic of the pattern, and reusing that generic logic, should make it easy to understand the tests *and* easy to write lots of independent tests; this should be consistent for all services

DrPyser (changes requested):
See comments

integration_tests/suite/test_http_interface.py
```diff
@@ -143,6 +143,130 @@ def test_that_no_type_returns_400(self):
...
+            ('POST', f'{base_url}/0.1/groups'),
+            ('PUT', f'{base_url}/0.1/groups/{test_group["uuid"]}'),
+        ]
+
+        for url in urls:
+            response = self._make_http_request(url[0], url[1], '')
+            assert response.status_code == 400, f'Error with url: {url}'
+
+            response = self._make_http_request(url[0], url[1], None)
+            assert response.status_code == 400, f'Error with url: {url}'
```
DrPyser:
```suggestion
        for method, url in urls:
            response = self._make_http_request(method, url, '')
            assert response.status_code == 400, f'Error with {method} {url}'

            response = self._make_http_request(method, url, None)
            assert response.status_code == 400, f'Error with {method} {url}'
```

rbienvenu0: Updated

integration_tests/suite/test_http_interface.py
```diff
@@ -143,6 +143,130 @@ def test_that_no_type_returns_400(self):
...
+        return call(
+            url,
+            headers=req_headers,
+            data=body,
+            verify=False,
+        )
+
+    @fixtures.http.group()
+    def test_empty_body_for_group_requests(self, test_group):
+        base_url = f'http://127.0.0.1:{self.auth_port()}'
```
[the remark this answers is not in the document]
rbienvenu0: Updated

integration_tests/suite/test_http_interface.py
rbienvenu0: Updated

integration_tests/suite/test_http_interface.py
```diff
@@ -143,6 +143,65 @@ def test_that_no_type_returns_400(self):
 
         assert_that(response.status_code, equal_to(400))
 
+    @fixtures.http.group()
+    def test_empty_body_for_group_requests(self, test_group):
```
DrPyser: the test names don't follow the pattern of existing tests.

Should be `test_empty_body_for_group_returns_400`.

Also, as discussed, different operations (POST/PUT) deserve separate tests, since they may fail independently and for different reasons.

Ideally, it would be `test_create_group_empty_body_returns_400` and `test_update_group_empty_body_returns_400`.

Note how a fixture needs to be created for the `PUT` test, but is irrelevant for the `POST` test. If `POST` is broken, the fixture gets created for nothing, which delays the result. And if an issue occurs in the fixture creation, the `POST` test is impacted when it shouldn't be.

integration_tests/suite/test_http_interface.py
```diff
@@ -143,6 +143,81 @@ def test_that_no_type_returns_400(self):
...
+    def test_that_empty_body_for_request_returns_400(self):
+        urls = [
+            ('PATCH', 'config'),
+            ('POST', 'external/google/config'),
+            ('PUT', 'external/google/config'),
+            ('POST', 'saml/sso'),
+            ('PUT', 'idp/saml/users'),
+            ('PUT', 'backends/ldap'),
+        ]
+        self.assert_empty_body_returns_400(urls)
```
DrPyser: please split unrelated endpoints in separate tests.

wazo_auth/plugins/http/password_reset/http.py
```diff
@@ -74,7 +74,7 @@ def post(self):
             raise Unauthorized(token_id)
 
         try:
-            args = PasswordResetPostParameters().load(request.get_json())
+            args = PasswordResetPostParameters().load(request.get_json(force=True))
```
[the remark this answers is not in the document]
rbienvenu0: Updated

DrPyser (dismissed):
LGTM, well done.

DrPyser (dismissed):
take 2

PR #436, opened 2026-03-23
[ITEM-209] WP-1983-force-auth-cli-password

fblackburn1:
Why: after a VM migration or similar event, the wazo-auth-cli config
file can disappear while the user still exists in the DB. Running
wazo-auth-bootstrap complete would fail with "user already exists with
different credential" and leave the system unrecoverable without manual
DB intervention.

Reset the DB password and rewrite the config file instead of raising.

---

> [!NOTE]
> **Medium Risk**
> Changes `wazo-auth-bootstrap complete` behavior to forcibly reset an existing internal user’s password and revoke sessions, which impacts authentication state and could lock out unexpected users if misapplied. Logic is small but touches credential and session handling.
> 
> **Overview**
> Makes `wazo-auth-bootstrap complete` resilient to missing `wazo-auth-cli` credentials by **resetting the existing internal user password (and revoking its sessions) or creating the user**, then rewriting the CLI config instead of failing with a credential mismatch.
> 
> Adds a new `bootstrap` integration-test asset plus `test_bootstrap.py` to verify `complete` is idempotent and that deleting the CLI config triggers password reset and produces working credentials.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b8625e0b1ae48922b8158003afe9a307b2584a50. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/integration_tests/assets/docker-compose.bootstrap.override.yml
+++ b/integration_tests/assets/docker-compose.bootstrap.override.yml
@@ -1,0 +1,8 @@
+services:
+  sync:
+    depends_on:
+      - auth
+      - postgres
+      - rabbitmq
+    environment:
+      TARGETS: "auth:9497 postgres:5432 rabbitmq:5672"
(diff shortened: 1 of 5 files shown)


bloom1 (approved):
LGTM. Thank you!

PR #440, opened 2026-04-29
typing

DrPyser:
- **token dao: fix 3-level relative import**
- **typing: targeted config**
- **wazo_auth.database.queries.token: typing**

---

> [!NOTE]
> **Medium Risk**
> Mostly typing/tooling, but it changes DAO runtime behavior (NULL metadata normalization and `.one()` tenant lookup) which could raise in unexpected DB states or affect cleanup/notification jobs.
> 
> **Overview**
> Adds **mypy** to pre-commit and introduces a strict-ish `pyproject.toml` mypy configuration, initially enabling checks only for `wazo_auth.database.queries.token` (global `ignore_errors=true`).
> 
> Refactors `TokenDAO` (`wazo_auth/database/queries/token.py`) to use `TypedDict`-based return types, fixes the import to `from wazo_auth import exceptions`, and adjusts query/result shaping so expiring-token rows always return `metadata` as `{}` when the DB value is NULL (covered by a new integration test). It also tightens default-tenant lookup from `.first()` to `.one()` and adds runtime asserts after flush for UUIDs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99d7cd22b9e904864dc9144e1aeaa6255601004b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,11 @@ repos:
     rev: 1.1.1
     hooks:
       - id: wazo-copyright-check
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.19.1
+    hooks:
+      - id: mypy
+        language_version: '3.11'
+        additional_dependencies:
+          - 'sqlalchemy==1.4.46'
+          - 'sqlalchemy2-stubs'
--- a/integration_tests/suite/test_db_token.py
+++ b/integration_tests/suite/test_db_token.py
@@ -1,4 +1,4 @@
-# Copyright 2019-2024 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2019-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import time
@@ -164,6 +164,29 @@ class TestTokenDAO(base.DAOTestCase):
         assert_that(
             db_tokens,
             empty(),
+        )
+
+    def test_expiring_query_returns_empty_dict_when_metadata_is_null(self):
+        session = models.Session(tenant_uuid=self.top_tenant_uuid)
+        self.session.add(session)
+        self.session.flush()
+        token = models.Token(
+            auth_id='test',
+            session_uuid=session.uuid,
+            issued_t=int(time.time()),
+            expire_t=int(time.time()) - 1,
+            metadata_=None,
+        )
+        self.session.add(token)
+        self.session.flush()
+
+        tokens, _ = next(
+            self._token_dao.get_tokens_and_sessions_about_to_expire(0, batch_size=10)
+        )
+
+        assert_that(
+            tokens,
+            has_items(has_entries(uuid=token.uuid, metadata=equal_to({}))),
         )
 
     @fixtures.db.token(expiration=0)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,3 +20,32 @@ py_version = 311
 
 [tool.pytest]
 testpaths = ["wazo_auth"]
+
+[tool.mypy]
+python_version = "3.11"
+plugins = ["sqlalchemy.ext.mypy.plugin"]
+follow_imports = "silent"
+ignore_missing_imports = true
+show_error_codes = true
+check_untyped_defs = true
+warn_unused_configs = true
+disallow_untyped_calls = true
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
+warn_unused_ignores = true
+strict_equality = true
+extra_checks = true
+no_warn_no_return = true
+
+# avoid conflicts between modules of the same name e.g. setup.py
+explicit_package_bases = true
+
+# opt-in mode, disable when everything is sufficiently typed
+ignore_errors = true
+
+# extend as modules are typed, eventually remove or replace with exclusions
+[[tool.mypy.overrides]]
+module = [
+    "wazo_auth.database.queries.token",
+]
+ignore_errors = false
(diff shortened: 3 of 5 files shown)


fblackburn1 (commented):
# Code Review

## Overview
Bootstrap mypy in pre-commit (Python 3.11, SQLAlchemy 1.4 plugin/stubs) with an opt-in `[tool.mypy]` config (`ignore_errors = true` globally, overridden for `wazo_auth.database.queries.token`). The DAO file is then made strict-mypy-clean: `TypedDict`s for token/session shapes, full signatures on every method, `_asdict()` replaced by explicit dict construction, an absolute import of `exceptions`, and a post-flush non-null assertion in `create()`.

Risk is low — no behavior changes outside `TokenDAO.create` (added assert) and the TypedDict construction (semantically identical to dict literals at runtime).

## Strengths
- **Incremental rollout pattern is right.** Global `ignore_errors = true` + module-targeted override is the standard way to introduce mypy without a flag day. The comment "extend as modules are typed" makes the intent obvious.
- **Strict knobs from day one** (`disallow_untyped_defs`, `check_untyped_defs`, `strict_equality`, `extra_checks`, `warn_unused_ignores`) — better to start strict on the small surface than loosen later.
- **`TypedDict` for record shapes** documents the contract that callers like `TokenService.remove_token` rely on (`session['uuid']`, `token['auth_id']`).
- **Import cleanup** (`from wazo_auth import exceptions` instead of `from ... import exceptions`) — three-level relatives are brittle and the new form is unambiguous.

## Issues / Suggestions

### 1. `assert` for non-null UUIDs is fragile (token.py:80)
```python
self.session.flush()
assert token.uuid is not None and token.session_uuid is not None
return token.uuid, token.session_uuid
```
`assert` is stripped under `python -O`. If wazo-auth is ever run with optimizations, the function silently returns `(None, None)` and downstream sees a corrupted token. Prefer:
```python
if token.uuid is None or token.session_uuid is None:
    raise RuntimeError(\"Token UUID not assigned after flush\")
```
or `typing.cast(str, token.uuid)` if the goal is purely to satisfy mypy.

### 2. `delete()` return type is awkward (token.py:97-130)
```python
def delete(self, token_uuid: str) -> tuple[
    TokenIdentity | dict[str, Any],
    SessionIdentity | dict[str, Any],
]:
```
The `| dict[str, Any]` branch exists only to model the empty-`{}` not-found case, but `TypedDict` is already a `dict` subclass at runtime, so this union doesn't actually help mypy narrow anything — callers will still need to check truthiness before indexing. The TODO already acknowledges this. Consider doing it now:
```python
def delete(self, token_uuid: str) -> tuple[TokenIdentity, SessionIdentity] | tuple[None, None]:
```
or `Optional[tuple[TokenIdentity, SessionIdentity | None]]`. The single caller (`services/token.py:203-209`) only checks `if not session: return`, so an `Optional` is a clean fit.

### 3. `metadata: dict[str, Any]` may not match reality (token.py:39, 163-169)
`TokenModel.metadata_` is `Column(Text)`, which can be NULL. In `_get_tokens_and_sessions_by_expiration` the value is `cast(metadata_, JSONB).label('metadata')` — JSONB on a NULL text column yields `None`, not `{}`. The TypedDict says non-optional `dict[str, Any]`. In `get()`, the explicit `json.loads(...) if token.metadata_ else {}` guards this, but the expiring-row path does not. Either:
- Coalesce in SQL (`func.coalesce(cast(...), cast('{}', JSONB))`), or
- Type the field as `dict[str, Any] | None` and have callers handle.

Same applies for `pbx_user_uuid`, `xivo_uuid`, `remote_addr`, `user_agent`, `refresh_token_uuid` — already typed `| None`, good.

### 4. `_get_default_tenant_uuid` will crash on empty result (token.py:84-86)
```python
return self.session.query(Tenant).filter(filter_).first().uuid
```
`first()` returns `Optional[Tenant]`. Annotated `-> str` but `.uuid` on `None` raises `AttributeError`. Mypy with `sqlalchemy2-stubs` should flag this — worth checking why it didn't, or whether this method has callers that actually exercise the empty case.

### 5. SQLAlchemy version drift risk (.pre-commit-config.yaml)
`sqlalchemy==1.4.46` is hard-coded in `additional_dependencies`. If the project bumps SQLAlchemy in `setup.py` and forgets here, mypy stubs will drift silently and produce false positives/negatives. Not blocking, but a brief note in the README or a CI check would help.

### 6. `TokenRecord` extends `TokenIdentity` — `get()` returns more than callers want
`services/token.py:218` does `token_data.pop('uuid')` and unpacks the rest as kwargs into `Token(...)`. With `TypedDict`, `pop` works at runtime but is structurally wrong (removing a required key). Not a bug today, but when that file is added to the mypy override it will surface. Consider returning the `Token` object directly from the DAO, or splitting the \"wire format\" from the \"domain object\" cleanly.

## Test Coverage
- No new tests; appropriate for a typing-only PR.
- The one runtime change worth a smoke test is the post-flush `assert` in `create()` — existing token-creation integration tests should cover the happy path; verify they pass under `tox -e integration`.

## Security
- No surface-area changes. The DAO was already parameterized via SQLAlchemy.

## Recommendation
Approve after addressing **(1)** at minimum (replace `assert` with an explicit raise) and ideally **(2)** (`Optional` return for `delete`) since that TypedDict-or-empty-dict union will be copy-pasted elsewhere as more modules opt in. Items 3–6 can be follow-ups.

fblackburn1 (approved):

PR #445, opened 2026-06-16
ITEM-404 integration_tests: use shared wazo_test_helpers.pytest_asset plugin

ja0x:
Depends-On: https://github.com/wazo-platform/wazo-test-helpers/pull/127

## What

Replaces the per-suite asset lifecycle boilerplate in the integration suite
(collection ordering, eager marker-boundary teardown, teardown-failure
reporting, `mark_logs`) with the shared `wazo_test_helpers.pytest_asset` plugin
introduced in wazo-platform/wazo-test-helpers#127.

The conftest now only declares its assets via `asset_fixture()`, opts into log
markers via `enable_mark_logs_fixture()`, and calls `register()` from its own
`pytest_configure` to activate the hooks.

### DIFF
--- a/integration_tests/suite/conftest.py
+++ b/integration_tests/suite/conftest.py
@@ -1,72 +1,115 @@
 # Copyright 2021-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+import logging
+from collections.abc import Callable, Iterator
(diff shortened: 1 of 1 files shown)


ja0x:
recheck

DrPyser (approved):
LGTM, thanks

DrPyser:
see also https://github.com/wazo-platform/wazo-chatd/pull/203

PR #448, opened 2026-07-22
TASK-503 role-based-controller

fblackburn1:
Notes
- Options starting by `update_*_on_startup` should be analyzed if still necessary
- `ServiceDiscoveryComponent` and db startup lock helpers could be extracted to xivo-lib-python
- internal service need to use nginx proxy by default

---

> [!NOTE]
> **High Risk**
> Changes authentication service startup, migrations, periodic token/session cleanup, and deployment (systemd, nginx, removed --http-worker); misconfiguration could leave no scheduler leader or block startup on lock contention.
> 
> **Overview**
> **wazo-auth can run as multiple coordinated processes or replicas** by splitting work into **`api`**, **`scheduler`**, and **`init`** roles (config map + repeatable `--role` CLI). Default behavior stays one process with all roles. **`--http-worker` is removed**; API-only workers use `--role api` (systemd template `wazo-auth-worker@<port>`).
> 
> **Cross-instance safety** uses PostgreSQL advisory locks: **`init`** one-shots (policy update, bootstrap, IdP checks) and Alembic upgrades run under a **startup lock**; **expired token cleanup** runs only on a **single scheduler leader** elected with a dedicated lock (failover and release on repeated failures).
> 
> **Operational fixes:** `--listen-port` now sets `rest_api.port` (not listen address); **`service_discovery.advertise_port`** defaults to the real listen port; **`wazo-auth-wait`** accepts `--port` and `--timeout`. **SO_REUSEPORT / shared-port HTTP workers** are dropped in favor of distinct ports behind a shared **nginx `wazo-auth` upstream**.
> 
> Docs (`contribs/scaling/README.md`, changelog) and integration tests cover **two-replica cluster** behavior (shared tokens, one scheduler leader, cold-start lock).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 315b6b473c8972a6c5bfdef928968a27c71ee13e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


debian/wazo-auth.wazo-auth-worker@.service
```diff
@@ -1,15 +1,14 @@
...
 After=network.target postgresql.service wazo-auth.service
 PartOf=wazo-auth.service
 StartLimitBurst=15
 StartLimitIntervalSec=150
 
 [Service]
-ExecStartPre=/usr/bin/wazo-auth-wait
-ExecStart=/usr/bin/wazo-auth --http-worker
+ExecStart=/usr/bin/wazo-auth --role api --listen-port %i
+ExecStartPost=/usr/bin/wazo-auth-wait --port %i
```
ja0x:
```suggestion
ExecStartPost=-/usr/bin/wazo-auth-wait --port %i --timeout 300
TimeoutStartSec=300
```
Consider adding a `--timeout`  to wazo-auth-wait and/or setting the value to the db_ready (300s) + top tenant check (300s) = 600s. The current value is hard-coded to 60 seconds. In the case where a migration is slow and the service takes more than > 60s to boot, systemd will SIGTERM the workers and will set the status to `failed`

This also applies to the main "wazo-auth.service"

wazo_auth/config.py
```diff
@@ -9,10 +9,12 @@
 
 TWO_HOURS = 60 * 60 * 2
 _DEFAULT_HTTP_PORT = 9497
+VALID_ROLES = ('api', 'scheduler', 'init')
 _DEFAULT_CONFIG = {
     'user': 'wazo-auth',
     'config_file': '/etc/wazo-auth/config.yml',
     'extra_config_files': '/etc/wazo-auth/conf.d',
+    'roles': list(VALID_ROLES),
```
ja0x: The config file hierarchy for lists is cumulative instead of overriding. 
Example is having an override file (conf.d) for plugins, you want a new plugin to be added to the others, not override the plugin list.

Using config, an override of `roles: ['api']` in a conf.d file will be represented as [`api`, `scheduler`, `init`, `api`] spawning a wazo-auth with all roles instead of just `api`.

The CLI args works because it is a different mechanism than the config file.
The role key will need to be interpreted differently than the default config chain_map

wazo_auth/database/helpers.py
```diff
@@ -78,3 +88,144 @@ def db_ready(timeout):
...
     with get_db_session().get_bind().connect() as conn:
         conn.execute(text('SELECT 1'))
+
+
+class StartupLockInterrupted(Exception):
+    pass
+
+
+@contextmanager
+def startup_lock(engine, stop_event=None):
```
ja0x:
```suggestion
def startup_lock(engine, timeout, stop_event=None):
```
This method has no timeout/deadline, a hung upgrade will freeze startup forever without any logging. At minimum, there should be a log line describing who holds the lock and for how long after a delay so operators are not blind on silent failure

wazo_auth/token.py
```diff
@@ -123,47 +127,85 @@ def __init__(self, config, dao, bus_publisher, saml_service):
...
-            self._tombstone.clear()
+        if self._cleanup_interval < 1:
+            return
+        if not self._thread.is_alive():
+            # the signal handler can call stop() before run() reached
+            # start(): joining a never-started thread raises
+            return
+        self._tombstone.set()
+        self._thread.join()
+        self._tombstone.clear()
```
ja0x:
```suggestion
        if self._cleanup_interval < 1:
            return
        
        self._tombstone.set()
        if self._thread.is_alive():
            self._thread.join()
```
`self._loop` checks the `_tombstone`, in case of race condition, where `SIGTERM` was sent before the `start()` (so stop() called before start()), the thread would start and never join. This ordering makes sure to set the condition to stop _loop

wazo_auth/config.py
```diff
@@ -203,18 +207,38 @@ def _parse_cli_args(argv):
...
     if parsed_args.db_upgrade_on_startup:
         result['db_upgrade_on_startup'] = parsed_args.db_upgrade_on_startup
-    result['http_worker'] = parsed_args.http_worker
+    if parsed_args.roles:
+        result['roles'] = parsed_args.roles
 
     return result
 
 
+def _normalize_roles(roles):
```
ja0x, line 216: And invalid role will exit wazo-auth without logging, because the logger are configured after this method raises. The journal/systemd will show the traceback, but logs won't. 

This is a chicken and egg problem: `setup_logging` requires config values to run, but `get_config` can raise and wont be able to write logs. Possibly a try/except could help

Also related, by dropping exitcode 78 on misconfiguration, systemd will restart the service until the it reaches the restart limit, but a misconfiguration is not a recoverable error.  The exitcode 78 (defined in service file) prevents restarting a service on that error

CHANGELOG.md
```diff
@@ -2,10 +2,20 @@
 
 ## 26.08
 
-* Added `--http-worker` flag to start HTTP-only wazo-auth instance
-* New `rest_api.min_threads` option: threads kept ready at all times.
```
ja0x, line 6: min_thread should still be part of the changelog

CHANGELOG.md
```diff
@@ -2,10 +2,20 @@
...
-  `max_threads` is now a ceiling the pool grows to under load, not a fixed
-  thread count.
+* wazo-auth can be scaled to multiple processes or replicas:
```
ja0x, line 5: Will need to be moved to 26.09

wazo_auth/token.py
```diff
@@ -123,47 +127,85 @@ def __init__(self, config, dao, bus_publisher, saml_service):
...
+            self._leader_lock.release()
+
+    def _hold_leadership(self):
+        try:
+            return self._leader_lock.hold()
+        except Exception:
+            logger.warning(
+                'unable to determine scheduler leadership',
+                exc_info=self._debug,
+            )
```
ja0x, line 191:
```suggestion
            logger.warning(
                'unable to determine scheduler leadership: %s',
                exc, 
                exc_info=self._debug,
            )
```
Without debug, the error message provides no context of why `_hold_leadership` failed. Could add a stringified version of the exception instead of full traceback?

This applies to a few places in this file

ja0x (changes requested):
Tested locally, was able to start workers correctly and it seems to work. 
There are a few issues to validate/fix mostly related to startup/shutdown, splitting work across multiple processes is bound to have a lot of edge cases and harder to synchronize

wazo_auth/config.py
fblackburn1: This behavior is only between config files (not the default config dict)
Since the default `config.yml` is empty, then only two different `conf.d` could accumulate, which is a non-sense to have two different plugins that update this value since they don't even know each other and their priorities between them.

Anyway, I can follow the config convention for enabled_plugins and use dict instead `{'api': True, ...}`

debian/wazo-auth.wazo-auth-worker@.service
fblackburn1: Fixed in 760ab32c. `wazo-auth-wait` now accepts `--timeout`, and both units use `--timeout 600` + `TimeoutStartSec=600` (600 = db_ready 300 + top-tenant 300, per your math — `TimeoutStartSec` is needed alongside because systemd's default 90s covers `ExecStartPost`). I kept `ExecStartPost` without the `-` prefix so a readiness failure still fails the unit and triggers `Restart=on-failure`.

wazo_auth/config.py
fblackburn1: Fixed in 44a0e76c. `roles` is now a map of role name to boolean, like the `enabled_*_plugins` keys, so conf.d overrides merge per-key instead of accumulating. An api-only override is `roles: {scheduler: false, init: false}`; `--role` still replaces the whole set. The shipped config.yml now enables all three roles explicitly.

wazo_auth/database/helpers.py
fblackburn1: Fixed in 80d4f828. `startup_lock(engine, timeout, stop_event=None)` raises `StartupLockTimeout` past the deadline (both call sites reuse `db_connect_retry_timeout_seconds`, no new config key), and while blocked it logs the holder — pid, client address, application name, backend age from `pg_locks` ⋈ `pg_stat_activity` — immediately and every 30s. On timeout the process exits nonzero and `Restart=on-failure` retries, so a legitimately long migration on another node delays startup rather than failing it permanently.

wazo_auth/token.py
fblackburn1: Fixed in ddfb4e15, suggestion taken as-is. Note the trailing `tombstone.clear()` also had to go: keeping it would re-arm the leak (set → no-op join → clear → late `start()` runs forever). Added a regression test for the exact stop-then-start race.

wazo_auth/config.py
fblackburn1: Partially fixed in 1e55eb29. A config validation error is now caught, printed as a one-liner to stderr, and exits 78 (`os.EX_CONFIG`); `RestartPreventExitStatus=78` is restored on the worker unit and added to the main unit — previously it was declared but nothing ever actually exited 78. On the log-file half we kept the platform behavior: `get_config`-before-`setup_logging` is the pattern in every wazo service and journald captures stderr, so diverging only in wazo-auth didn't seem worth it.

CHANGELOG.md
fblackburn1: Restored in e5bab7db — along with the `--http-worker` bullet: both shipped in 26.08, so their entries stay under 26.08.

CHANGELOG.md
fblackburn1: Done in e5bab7db — new entries moved to a 26.09 section, with an explicit note that `--http-worker` was removed and replaced by `--role api`.

wazo_auth/token.py
fblackburn1: Fixed in cf676a48, applied to all three debug-gated warnings in the file (`_hold_leadership`, `_run_once`, `_purge_expired_sessions`): the stringified exception is always in the message, the full traceback stays behind debug.

wazo_auth/http_server.py
ja0x, line 108: CoreRestApi has the same race condition/issue TokenRenewer had and fixed in the stop. If stop() (from SIGTERM) is called before the start(), there's a window where it can still start and will prevent wazo-auth from stopping correctly, prompting systemd to SIGKILL.

Fix: mirror TokenRenewer, add a _stopped event that is set in stop(). Then in run() check the flag before starting to close the window

(picked up by Claude)

wazo_auth/controller.py
```diff
@@ -308,178 +315,116 @@ def __init__(self, config):
...
+                if time.monotonic() >= end_time:
+                    raise
                 logger.warning(
-                    'Authentication method %s is in use but is not available', method
+                    'the database is not initialized yet (%s), retrying...', e
                 )
+                Session.remove()
+                if self._stopped.wait(2):
+                    # shutdown requested during the wait
+                    return
```
ja0x, line 428:
```suggestion
        end_time = time.monotonic() + timeout
        first_failure = True
        while True:
            try:
                http.init_top_tenant(self.dao)
                return
            except Exception as e:
                if time.monotonic() >= end_time:
                    raise
                logger.warning(
                    'the database is not initialized yet (%s), retrying...', 
                    e,
                    exc_info=first_failure,
                )
                first_failure = False
                Session.remove()
                if self._stopped.wait(2):
                    # shutdown requested during the wait
                    return
```
To help debugging/diagnostics on a failed initialization where the exception message is not enough context, the traceback could be printed for the first failure only (since it will be retries every 2 seconds afterwards

wazo_auth/main.py
```diff
@@ -20,33 +21,49 @@
...
             raise
 
-    worker_suffix = ' (http worker)' if config.get('http_worker') else ''
-    logger.info('Launching wazo-auth%s...', worker_suffix)
-
     controller = Controller(config)
     controller.run()
 
 
 def upgrade_db():
```
ja0x, line 64: `wazo-auth-upgrade-db` entrypoint is `upgrade_db`. The config does not have the same protection as `wazo-auth` entrypoint (try/except around get_config). A broken roles would crash the entrypoint

To confirm, that might be voluntary behavior based on this being run by apt command

wazo_auth/token.py
```diff
@@ -123,47 +127,84 @@ def __init__(self, config, dao, bus_publisher, saml_service):
...
+            Session.close()
+            self._consecutive_failures += 1
+            if self._consecutive_failures >= self.MAX_CONSECUTIVE_FAILURES:
+                logger.warning(
+                    'releasing the scheduler leader lock after %d '
+                    'consecutive failures',
+                    self._consecutive_failures,
+                )
+                self._leader_lock.release()
```
ja0x, line 215: Nothing gates the same renewer from fetching the lock again on next iteration and it will probably win. 

For single host, this is probably the behavior we want, but in a "cloud" environment (>1 scheduler role or a respawning healthy scheduler role) the same process might keep the lock the whole time. A small backoff tick counter could address that by allowing a grace period before the same failing process can get leadership again

ja0x (changes requested):

wazo_auth/http_server.py
fblackburn1: Good catch, but this is pre-existing rather than introduced here: master has the identical `if self.server:` no-op (and no `_stopped` gate at all before `rest_api.run()`), and the same pattern exists in wazo-dird, wazo-calld, wazo-confd, etc. Since all services should get the same fix, it is tracked in the scaling backlog to fix platform-wide: https://app.notion.com/p/3b4aeaae6d2a8153ab50d787b088cdfb

wazo_auth/controller.py
fblackburn1: Fixed in bdad0a34, suggestion taken: the full traceback is logged on the first failure only (`exc_info=first_failure`), retries stay one-liners. The rest of the suggestion (deadline, `_stopped.wait`) was already in place.

wazo_auth/main.py
fblackburn1: Fixed in 140e9402. The crash-on-bad-config part is intentional — the postinst must fail loudly on a broken config either way — but the raw traceback was an oversight: `upgrade_db` now has the same try/except as `main()`, printing a one-liner to stderr and exiting 78 (the exit code has no functional effect for apt, kept purely for consistency between the two entrypoints).

wazo_auth/token.py
fblackburn1: Fixed in 315b6b47 with the backoff tick counter you suggested: after a failure-triggered release the instance sits out `FAILURE_BACKOFF_TICKS = 2` ticks before contending again. Two ticks rather than one because the sick leader releases mid-tick and re-contends after `interval - elapsed`, while a healthy instance whose tick just missed the release waits a full interval — with one tick the sick one can still win the race. Single-host trade-off: cleanup pauses two extra intervals before the same process retakes the lock, which just delays token purging by a couple of minutes.

ja0x (approved):

fblackburn1:
regate

fblackburn1:
regate :clown_face:

PR #450, opened 2026-08-12
Reuse nginx upstream connections (keepalive) to wazo-auth

fblackburn1:
TASK-506 — split out of TASK-504 (#449), which deliberately kept nginx's default connection behaviour.

## What

- `keepalive 8` in the `wazo-auth` upstream block, plus `proxy_http_version 1.1` and `proxy_set_header Connection ""` in `wazo-auth-shared.conf`. The three directives are one feature: on nginx 1.22 (bookworm) the proxy defaults are HTTP/1.0 + `Connection: close`, which makes `keepalive` alone inert. nginx >= 1.30 makes the two proxy directives the default; they stay explicit so pooling does not depend on the distro nginx. Both entry points (https and internal) pool, since they share the upstream block.
- Raise cheroot's `keep_alive_conn_limit` from its class default of 10 to 128 so the pool survives between bursts: nginx parks up to 8 idle connections per worker process, one worker per core. Deliberately not documented in `config.yml`; `rest_api.keep_alive_conn_limit` can still be overridden from conf.d.

## Measured

nginx 1.22 + cheroot 9.0.0 stub serving the real location files:

| scenario | requests | backend connections |
| --- | --- | --- |
| baseline (master) | 2000 sequential | 1932 |
| this branch | 2000 sequential | 2 |
| this branch, limit 10 via conf.d | 4000 @ concurrency 40 | 240 |
| this branch, limit 128 | 4000 @ concurrency 40 | 40 |

Also validated on a real stack with the same with/without comparison.

## Notes

- An nginx reload or a wazo-auth restart invalidates pooled connections; nginx retries on a fresh one, so at most a transient error during upgrades.
- Each parked connection holds one file descriptor in cheroot's selector (up to 128).
- Client-side reuse (`wazo-lib-rest-client`) is PRODUCT-259, not this PR.

---

> [!NOTE]
> **Low Risk**
> Connection-pooling and tuning only; no changes to auth, tokens, or request handling logic.
> 
> **Overview**
> **Enables HTTP keep-alive between nginx and wazo-auth** so repeated API calls can reuse upstream TCP connections instead of opening a new one per request.
> 
> nginx gets an upstream `keepalive` pool (8 idle connections per worker) plus shared proxy settings (`HTTP/1.1`, empty `Connection` header) required for that reuse. On the app side, **wazo-auth’s Cheroot server** now honors a new `rest_api.keep_alive_conn_limit` default (**128**) so it can hold more concurrent client keep-alive connections from the proxy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf81d9606b039188b53bdbce2cace18ae90ff9cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


etc/nginx/conf.d/wazo-auth-upstream.conf
```diff
@@ -4,4 +4,8 @@
 #     server 127.0.0.1:19497;
 upstream wazo-auth {
     server 127.0.0.1:9497;
+
+    # Idle connections parked per worker process. Inert without the
+    # proxy directives in wazo-auth-shared.conf.
+    keepalive 8;
```
ja0x:
```suggestion
    keepalive 8;
    keepalive_timeout 5;
```
By default nginx holds connections for 60s while cheroot expires connections after 10 seconds. With the default, a "dead" connection can get picked and return 502 on non-idempotent requests (e.g: POST). A 5 seconds keepalive timeout would prevent that case at the cost of connection churning on low activity (same behavior as today)

fblackburn1: Applied, thanks

fblackburn1: Also raised `keepalive` from 8 to 32 in the same commit, so we match the
default nginx >= 1.30 applies anyway — one less thing that silently changes
under us on a distro bump.

That does move the backend-side ceiling: at 32 per worker with one worker per
core, `keep_alive_conn_limit: 128` only covers 4 cores, past which cheroot
stops parking and answers `Connection: close`. Race-free, but it caps pooling
exactly where the previous commit set out to remove the cap — so I bumped it to
256 (8 cores) and updated that commit's rationale.

